PR #19, opened 2020-04-17
chore | Add start script for running the examples via parcel

richiemccoll:
I've added a start script so that the examples (which have parcel as a devDependency) can be run locally.


brainkim:
Thank you!

PR #21, opened 2020-04-17
Fix small error in example

nicosommi:


brainkim:
Oof good catch! Thank you!

PR #22, opened 2020-04-17
Change background to gray

SharpSeeEr:
For the love of all that is holy, change the background from that blue! 😛

### DIFF
--- a/website/src/index.css
+++ b/website/src/index.css
@@ -5,7 +5,7 @@
 }
 
 body {
-	background-color: #113576; 
+	background-color: #242424; 
 	color: #f5f9ff;
 	font-family: sans-serif;
 }


brainkim:
Yes I keep getting this feedback! I’ve been tweaking the colors and I think I’ve gotten something that is a little uhhh less offensive. Apparently, it’s not just the contrast of text/background but also the saturation? Like human eyes get tired from saturated colors which makes sense when you’re like looking for too long at a blue sky. Check out the site now and see if it’s less eye-hurting?

SharpSeeEr:
Definitely better!  But the dark blue still does weird things to my eyes.  However, I appreciate very much your responsiveness, and will be watching this project.  I had my doubts about this project but because of the positive feedback on Reddit I have decided I should keep an open mind!

Closing this PR.

brainkim:
Happy to hear! If you can put any of your concerns to prose, I’m all ears!

PR #35, opened 2020-04-18
feat(context): add settimeout that clears on unmount

JasonMatthewsDev:
Adds a `setTimeout` method to the base context that is cleared when the component is unmounted.

Definitely open to any feedback. Awesome project really looking forward to seeing this develop.

### DIFF
--- a/src/__tests__/index.tsx
+++ b/src/__tests__/index.tsx
@@ -1147,6 +1147,63 @@ describe("sync generator component", () => {
 			"Use for...of",
 		);
 	});
+	
+	describe("setTimeout", () => {
+		beforeAll(jest.useFakeTimers);
+		afterAll(jest.useRealTimers);
+		afterEach(jest.clearAllTimers);
+
+		test("set timeouts", () => {
+			function *TimeOut(this: Context): Generator<Element> {
+				let text = "foo";
+				this.setTimeout(() => { 
+					text = "bar";
+					this.refresh();
+				}, 100);
+				this.setTimeout(() => { 
+					text = "baz";
+					this.refresh();
+				}, 200);
+
+				for (const _ of this) {
+					yield <div>{text}</div>;
+				}
+			}
+		
+			renderer.render(<TimeOut />, document.body);
+			expect(document.body.innerHTML).toBe("<div>foo</div>");
+			jest.advanceTimersByTime(101);
+			expect(document.body.innerHTML).toBe("<div>bar</div>");
+			jest.advanceTimersByTime(101);
+			expect(document.body.innerHTML).toBe("<div>baz</div>");
+		});
+
+		test("cancel timeouts", () => {
+			let text = "foo"
+			function *TimeOut(this: Context): Generator<Element> {
+				this.setTimeout(() => { 
+					text = "bar";
+					this.refresh();
+				}, 100);
+				this.setTimeout(() => { 
+					text = "baz";
+					this.refresh();
+				}, 200);
+
+				for (const _ of this) {
+					yield <div>{text}</div>;
+				}
+			}
+		
+			renderer.render(<TimeOut />, document.body);
+			renderer.render(<div />, document.body);
+			expect(text).toBe("foo");
+			jest.advanceTimersByTime(101);
+			expect(text).toBe("foo");
+			jest.advanceTimersByTime(101);
+			expect(text).toBe("foo");
+		});
+	});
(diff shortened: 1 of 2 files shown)


joncfoo:
What are some use cases where this would be beneficial?

JasonMatthewsDev:
This PR is inspired from a conversation the author was having on Reddit: https://old.reddit.com/r/javascript/comments/g1zj87/crankjs_an_alternative_to_reactjs_with_built_in/fnl6iss/

It would be beneficial any time there was a possibility of a component being unmounted after setting 1 or more timeouts that should not be executed if the component is no longer mounted.

It's more of a convenience than anything else. Crank will track the timeout references and clear them on unmounting when using the context method instead of having to do it manually.

joncfoo:
Ah gotcha, thanks for providing context!

JasonMatthewsDev:
Np! I appreciate the question. I should've added the context to the original PR.

lygaret:
This seems like something you could implement as additional functions on `Context.prototype` (as a plugin, of sorts), if there was a way to run code on unmount. I tried to do something similar directly, but realized I didn't have a way to emulate a `finally` block.

eg.

```typescript
Context.prototype.setTimeout = function<T>(this: Context, timeout: number, cb: (T) => void) {
    const ref = setTimeout(cb, timeout);

    // a 'willUnmount' listener would let us cleanup per instance
    this.addEventListener('willUnmount', () => {
      clearTimeout(ref);
    });
}
```

My use-case was actually rxjs subscription management, but it's the same idea; given the ability to register callbacks with unmount, we should be able to just willy-nilly add context methods where they make sense.

I'm probably missing something, but it seems like only a little bit further than what's already here and doesn't special case anything.

ryhinchey:
I’m working on adding an `unmount` event that components can listen to.  It’s looking promising locally.  I’ll try and make a pr soon to get feedback

plusgut:
In this example is shown how to remove your listeners https://github.com/bikeshaving/crank/blob/master/examples/hackernews/src/index.jsx#L147

It seems like crank is calling throw on the generators, when it is unmounting. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/throw

monodop:
an unmount event would be pretty useful. I've been thinking about some kind of `addSideEffect` function where you could register a generator that acts kind of like a parallel generator. You could use it to receive prop updates, handle setup and unmounting logic, just like making a regular component except without yielding any elements. But it could still call this.refresh() to trigger rerenders.

Something like this maybe?

```jsx
function* mySetTimeoutExtension(this, {prop}, stateRef) {

    stateRef.value = {
        tickRate: 0,
        ticks = 0,
    }

    let timer = setInterval(() => {
        stateRef.value.ticks += stateRef.value.tickRate;
        this.refresh();
    }, 1000)
    let tickRate = 0;
    let currentProp = prop;

    try {
        for ({prop} of this) {

            if (currentProp != prop) {
                stateRef.value.tickRate++;
                this.refresh();
            }

        }
    }
    finally {
        clearInterval(timer);
    }

}

function* MyComponent(this, props) {

    let tickRateRef = this.addExtension(mySetTimeoutExtension, 0);
    while (true) {
        const {ticks, tickRate} = tickRate.value;
        yield <div>Ticks: {ticks}, TickRate: {tickRate}</div>
    }

}```

brainkim:
@JasonMatthewsDev Thank you for giving this a shot! I think the code as is looks good, and it’s kind of incredible to have people poking around in src/index.ts as so far it’s just been me working on it. Unfortunately, we might need a more general solution to Context extensions, as people in this PR have commented. You can imagine, for instance, having to add more methods for `setInterval` and `requestAnimationFrame`, and I’d like to figure out an approach. I have some ideas I just need to put it in words. I’ll make sure to keep you in the loop if you’re interested!

JasonMatthewsDev:
No problem at all. I'm completely okay with this just helping start those other conversations. I've been following along and have been playing with a generic solution on my end also.

brainkim:
@lygaret An earlier iteration of the library had custom Crank events defined via the EventTarget interface. I removed this because:
1. These events should not capture or bubble, and having all components be able to listen to lifecycle methods on children or parents felt like a recipe for disaster.
2. Events imply that the callback can be called more than once. I think any callback-based API should fire once per call (think `setTimeout` not `setInterval`).

PR #44, opened 2020-04-21
Fix/htmlattrs - escapeText styles also and handles false attrs correctly

fritx:
Hello @brainkim 
This library is awesome and creative, thanks for this !
I'll try to understand it in the next few days.

Here is my first PR on html attrs, thanks.

### DIFF
--- a/src/html.ts
+++ b/src/html.ts
@@ -50,7 +50,7 @@ function printAttrs(props: IntrinsicProps<string>): string {
 			case name === "innerHTML":
 				break;
 			case name === "style":
-				attrs.push(`style="${printStyle(value)}"`);
+				attrs.push(`style="${escapeText(printStyle(value))}"`);
 				break;
 			case typeof value === "string":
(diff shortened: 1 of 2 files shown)


src/html.ts
```diff
@@ -59,7 +59,7 @@ function printAttrs(props: IntrinsicProps<string>): string {
 				attrs.push(`${escapeText(name)}="${value}"`);
 				break;
 			case typeof value === "boolean":
```
brainkim, line 61: `case value === true:`

brainkim: Then we don’t have to have an if statement in the body. Thanks for catching this.

fritx: You're right 😆  Fixed.
Feel free to squash my commits or just ask me to do it if needed.

brainkim:
Thank you for the fixes! I will make sure to deploy this ASAP. We probably want a more robust test suite to test XSS vulnerabilities but I appreciate the test.

brainkim:
Deployed in v0.1.1

PR #45, opened 2020-04-21
Feat/umdbundle - supports umd-bundle with corresponding rollup.config and npm-script

fritx:
For possible umd usage, we need a umd bundle:
- supports umd-bundle with corresponding rollup.config and npm-script
- call build:umd in build cmd

Live demo, check it out ---> 
https://coldemo.github.io/#/playground/crank-async-generator.jsx
https://coldemo.github.io/#/playground/crank-todomvc.jsx

(I've published a temp npm-pkg named 'crank-umd' to get it work)
(It can be removed at some time if crank gets umd supported itself)

(codesandbox can be broken due to some network issues in some area somehow)
(for example, I personally never had luck to open any of them in the docs)

Similar merged PR: https://github.com/kelekexiao123/genjijs/pull/6

PS: I tried to patch the existing rollup.config but failed with

```plain
src/index.ts, src/dom.ts, src/html.ts → umd, cjs, esm...
[!] Error: UMD and IIFE output formats are not supported for code-splitting builds.
```

so I've added a new file: rollup.config.umd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+umd
 cjs
 esm
 lib
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "types": "esm/index.d.ts",
   "scripts": {
     "prebuild": "yarn run clean",
-    "build": "rollup -c rollup.config.js",
+    "build": "rollup -c rollup.config.js && npm run build:umd",
     "build:umd": "rollup -c rollup.config.umd.js",
     "clean": "shx rm -rf ./cjs ./esm",
     "lint": "eslint --ext .js,.jsx,.ts,.tsx .",
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "scripts": {
     "prebuild": "yarn run clean",
     "build": "rollup -c rollup.config.js",
+    "build:umd": "rollup -c rollup.config.umd.js",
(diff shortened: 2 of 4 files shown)


fritx:
Updated: Live demo, check it out
https://coldemo.github.io/#/playground/crank-async-generator.jsx
https://coldemo.github.io/#/playground/crank-todomvc.jsx   (new added)

brainkim:
Thank you for working on this! I don’t fully understand why coldemo requires UMD and not CJS? We can ship a UMD version, but I would want it just to be rolled up (haha) into the general build script for simplicity’s sake. In other words, rather than creating a separate rollup config for umd, just add it to the current rollup config.

Also, I’m very sorry to hear that CodeSandbox isn’t available in your country! I’ve found that CodeSandbox has been incredibly helpful for bootstrapping demos but we may want to consider a different solution if it’s blocked for the international audience.

brainkim:
Closing in favor of #108, which should be released on Monday in an 0.2 release.

1 question: does it matter if we leave umd modules untranspiled?

brainkim:
> [!] Error: UMD and IIFE output formats are not supported for code-splitting builds.
By the way, this error happens because rollup can’t do multiple files for umd.

brainkim:
UMD shipped with 0.2.0 https://unpkg.com/browse/@bikeshaving/crank@0.2.0/umd/

PR #48, opened 2020-04-22
this.addExtension()

monodop:
Hey guys, in response to [this comment](https://github.com/bikeshaving/crank/pull/35#issuecomment-617930877) on #35, I made this pull request to discuss some kind of generalized extension api for Context.

My goal is to allow you to lift related code from different parts of your component and into a single extension call. My previous attempts at doing this required putting code in the `finally` block of the calling method, something that could be forgotten. This method will instead allow you to create an async generator that can respond to prop updates and handle cleanup gracefully.

So far I don't have much, just a work in progress, but I'd love to hear feedback on this approach! 

TODO: 
- [ ] Clean up code and add documentation
- [ ] Add more tests
- [ ] Fix the tests I broke (not sure why - something to do with race conditions)
- [ ] Find a way to make props type-safe?
- [ ] Add a way to define initial state, so that the returned ref isn't possibly undefined
- [ ] Utilize the yields from the async generator returned by the extension?

As of the current version: here's how you might implement a setTimeout extension:

```jsx
function setTimeoutExtension(cb: () => void, ms: number) {
	return async function* extension(this: Context, props: {}, state: StateRef<State>): AsyncGenerator<State> {
		let completed = false;
		state.value = setTimeout(() => {
			cb();
		}, ms) as any;

		try {
			for await (props of this) {
				if (completed = true)
					break;
			}
		}
		finally {
			clearTimeout(state.value);
		}
	}
}
```

usage in a component:
```jsx
function* Component(this: Context): Generator<Element> {
	let text = "foo";
	let timer = this.addExtension(setTimeoutExtension(() => {
		text = "bar";
		this.refresh();
	}, 100));
	let timer2 = this.addExtension(setTimeoutExtension(() => {
		text = "baz";
		this.refresh();
	}, 200));
	let timer3 = this.addExtension(setTimeoutExtension(() => {
		text = "oh no";
		this.refresh();
	}, 300));

	for (const _ of this) {
		clearTimeout(timer3.value);
		yield (
			<div>{text}</div>
		)
	}
}
```

### DIFF
--- a/src/__tests__/index.tsx
+++ b/src/__tests__/index.tsx
@@ -8,6 +8,7 @@ import {
 	Context,
 	Element,
 	Fragment,
+	StateRef,
 } from "../index";
 import {renderer} from "../dom";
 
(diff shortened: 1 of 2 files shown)


monodop:
One issue I'm running into is that I can't quite get the extension iterators to run before the main component iterator. I think the issue might be that when I publish to the async iterator, it doesn't run until after I give control back, but we immediately start calling into `.run` and processing the sync iterator.

monodop:
I updated the code to support sync generator functions. I'm having a really hard time figuring out how to get the async generators to work properly, since there seems to be a timing issue with the js task queue. To get this working properly, we'll need to make sure that the extension generators/iterators get run first, followed by the component generators/iterators. @brainkim I might need some help/insight from you on what the best way to approach this would be, assuming you like this extension api 🙂 

I wrote a new test extension and I think that it's looking pretty good:

```jsx
type State = string;
function propComparer(propName: string) {
	return function* extension(this: Context, props: any, state: StateRef<State>): Generator<State, any, State> {
		let prev = props[propName];
		state.value = "no changes";

		for (props of this) {
			let current = props[propName];
			let message = current === prev ? "no changes" : `changed from '${prev}' to '${current}'`;
			prev = current;
			yield message;
		}
	}
}

function* Component(this: Context): Generator<Element> {

	let compareMessage = this.addExtension(propComparer("name"));

	for (const {name} of this) {
		debugger;
		yield (
			<div>Name: {name}. {compareMessage.value}</div>
		)
	}

}
```

monodop:
Just to add on to the previous message, here's the problem area (I think):

```jsx
	refresh(): MaybePromise<undefined> {
		if (this.unmounted) {
			return;
		}

		this.refreshCounter++;
		let publishes = this.publishes;
		this.publishes = [];
		
		// These publish functions are resolve functions, so the code waiting on them is put onto the task queue
		publishes.forEach(publish => publish(this.props!));
		
		// This gets run before the publish iterators are completed, which means that async extensions will run after sync components
		return this.run();
	}
```

wmadden:
You could also just dispatch an event from the context object to indicate the component was unmounted. It's already an `EventTarget`.

monodop:
@wmadden I was considering adding a bunch of lifecycle events but I wanted to avoid going down the `componentWillUpdate` `componentDidUpdate` `componentWillUnmount` etc route. However I do think if you just left it as an unmount event it might be pretty useful in some situations.

brainkim:
There’s been a lot of churn and you’ll likely have to do the PR from scratch, if you’re still interested in this feature. I apologize1 In the meantime I’ve added some helper methods to contexts (Context.prototype.schedule and Context.prototype.cleanup), which might be an alternative approach to the problem you were trying to solve here.

monodop:
I haven't had a chance to look at the new helper methods yet - been kinda deep into some other stuff the last couple months. I'll definitely check it out when I have some time!

Issue #49, opened 2020-04-22
Typescript support for props in Component type


monodop:
This library does a decent job with typescript, but it'd be really nice to have some support for strongly typing props in the `Component` type, and it might also be nice to have the same in `Context`.  One use-case for this is to be able to use a `Component<MyExpectedProps>` in a function signature or config option. Another use-case would be defining type-safe higher order components. I can't see a way to get the prop types to propagate down properly as things are, but I could just be missing something.

Here's a concrete example of something I might want to do. I know I could pass the thing down with `this.set` but then I still don't get type-safety. With this method I could also require every route to have the prop.
```tsx
import {createElement, Context, Component} from '@bikeshaving/crank';

export type RouteProps = {
    currentRoute: Route;
}

export type Route = {
    name: string;
    index: Component<RouteProps>;
}

export type RouterProps = {
    routes: Route[];
};

function getCurrentPath(): string[] {
    return new URL(window.location.href)
        .hash
        .substring(1)
        .split('/')
        .map(p => p.trim())
        .filter(p => p);
}

function getMatchingRoute(routes: Route[], currentPath: string[]): Route | null {
    const p = currentPath[0];
    const matches = routes.filter(r => r.name == p);
    if (matches.length > 0)
        return matches[0];
    return null;
}

export function* Router(this: Context, {routes}: RouterProps) {
    let currentPath = getCurrentPath();
    const listener = () => {
        currentPath = getCurrentPath();
        this.refresh();
    }
    window.addEventListener('popstate', listener);
    try {
        while(true) {
            const route = getMatchingRoute(routes, currentPath);
            if (route) {
                const C = route.index;
                yield <C route={currentRoute} />;
            }
            else {
                yield <div>Not found</div>
            }
        }
    }
    finally {
        window.removeEventListener('popstate', listener);
    }
}
```

shirakaba:
This is all guesswork without having opened up a code editor, but I have a hunch:

## Current situation

In @brainkim 's TodoMVC example, [the following](https://github.com/bikeshaving/crank/blob/ce0d9ffa5e3c26f7023fc4bf361f11e77e113965/examples/todomvc-ts/src/index.tsx#L16) signature appears:

```tsx
function* Header(this: Context): Generator<Element> {
    // ...
}
```

[...]

PR #51, opened 2020-04-23
Improve typings to add safety for props

monodop:
Based on #49, I think I have everything right. Thoughts?

- [x] Add tests to prove types are working
- [ ] Cleanup commits
- [ ] Split tests to separate PR or Upgrade to TypeScript 3.9.0 https://github.com/microsoft/TypeScript/issues/37198

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "rollup-plugin-typescript2": "^0.27.0",
     "shx": "^0.3.2",
     "ts-jest": "^25.3.1",
-    "typescript": "^3.8.3"
+    "typescript": "3.9.1-rc"
   },
   "publishConfig": {
     "access": "public"
--- a/src/__tests__/types.tsx
+++ b/src/__tests__/types.tsx
@@ -1,0 +1,284 @@
+/** @jsx createElement */
+/* eslint @typescript-eslint/no-unused-vars: "off", jest/expect-expect: "off" */
+import {
+	createElement,
+	Component,
+	FunctionComponent,
+	Context,
+	GeneratorComponent,
+} from "..";
+
+describe("types", () => {
+	type MyProps = {
+		message: string;
+	};
+	let elem: any;
+
+	test("Not Components", () => {
+		// @ts-expect-error
+		const MyString: Component = "Hello";
+
+		// @ts-expect-error
+		const MyNumber: Component = 10;
+
+		// @ts-expect-error
+		const MyBoolean: Component = true;
+
+		// @ts-expect-error
+		const MyNull: Component = null;
+
+		// @ts-expect-error
+		const MyUndefined: Component = undefined;
+
+		// Not entirely sure why this one doesn't error anymore. seems that <a></a> returns
+		// any, while createElement('a') returns Element<any>.
+		// skip @ts-expect-error
+		// const MyElement: Component = <a></a>;
+
+		// @ts-expect-error
+		const NotComponent: Component = {};
+	});
+
+	test("Components", () => {
+		const MyFunctionComponent: Component<MyProps> = function (this, props) {
+			const ctx: Context<MyProps> = this;
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyFunctionComponent />;
+		elem = <MyFunctionComponent message={"message"} />;
+
+		const MyAsyncFunctionComponent: Component<MyProps> = async function (
+			this,
+			props,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyAsyncFunctionComponent />;
+		elem = <MyAsyncFunctionComponent message={"message"} />;
+
+		const MyGeneratorComponent: Component<MyProps> = function* (
+			this,
+			initialProps,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = initialProps.message;
+			// @ts-expect-error
+			let unexpected = initialProps.unexpected;
+
+			for (const newProps of this) {
+				let newMessage: string = initialProps.message;
+				// @ts-expect-error
+				let newUnexpected = newProps.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyGeneratorComponent />;
+		elem = <MyGeneratorComponent message={"message"} />;
+
+		const MyAsyncGeneratorComponent: Component<MyProps> = async function* (
+			this,
+			initialProps,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = initialProps.message;
+			// @ts-expect-error
+			let unexpected = initialProps.unexpected;
+
+			for await (const newProps of this) {
+				let newMessage: string = initialProps.message;
+				// @ts-expect-error
+				let newUnexpected = newProps.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyAsyncGeneratorComponent />;
+		elem = <MyAsyncGeneratorComponent message={"message"} />;
+	});
+
+	test("Function Components", () => {
+		const MyFunctionComponent: FunctionComponent<MyProps> = function (
+			this,
+			props,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyFunctionComponent />;
+		elem = <MyFunctionComponent message={"message"} />;
+
+		const MyAsyncFunctionComponent: FunctionComponent<MyProps> = async function (
+			this,
+			props,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyAsyncFunctionComponent />;
+		elem = <MyAsyncFunctionComponent message={"message"} />;
+
+		// @ts-expect-error
+		const MyGeneratorComponent: FunctionComponent<MyProps> = function* (
+			this,
+			props,
+		) {
+			yield <div></div>;
+		};
+		// @ts-expect-error
+		const MyAsyncGeneratorComponent: FunctionComponent<MyProps> = async function* (
+			this,
+			props,
+		) {
+			yield <div></div>;
+		};
+	});
+
+	test("Generator Components", () => {
+		const MyGeneratorComponent: GeneratorComponent<MyProps> = function* (
+			this,
+			initialProps,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = initialProps.message;
+			// @ts-expect-error
+			let unexpected = initialProps.unexpected;
+
+			for (const newProps of this) {
+				let newMessage: string = initialProps.message;
+				// @ts-expect-error
+				let newUnexpected = newProps.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyGeneratorComponent />;
+		elem = <MyGeneratorComponent message={"message"} />;
+
+		const MyAsyncGeneratorComponent: GeneratorComponent<MyProps> = async function* (
+			this,
+			initialProps,
+		) {
+			const ctx: Context<MyProps> = this;
+			let message: string = initialProps.message;
+			// @ts-expect-error
+			let unexpected = initialProps.unexpected;
+
+			for await (const newProps of this) {
+				let newMessage: string = initialProps.message;
+				// @ts-expect-error
+				let newUnexpected = newProps.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		};
+		// @ts-expect-error
+		elem = <MyAsyncGeneratorComponent />;
+		elem = <MyAsyncGeneratorComponent message={"message"} />;
+
+		// TODO: add ts-expect-error at some point in the future, I guess?
+		// This will not pass because the function is infered as any, and any matches Iterator<any, any, any>.
+		// Hopefully a later typescript version fixes this :(
+		const MyFunctionComponent: GeneratorComponent<MyProps> = function (
+			this,
+			props,
+		) {
+			return <div></div>;
+		};
+		// @ts-expect-error
+		const MyAsyncFunctionComponent: GeneratorComponent<MyProps> = async function (
+			this,
+			props,
+		) {
+			return <div></div>;
+		};
+	});
+
+	test("Loose Typings", () => {
+		function MyFunctionComponent(props: MyProps) {
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		}
+		// @ts-expect-error
+		elem = <MyFunctionComponent />;
+		elem = <MyFunctionComponent message={"message"} />;
+
+		async function MyAsyncFunctionComponent(props: MyProps) {
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			return <div></div>;
+		}
+		// @ts-expect-error
+		elem = <MyAsyncFunctionComponent />;
+		elem = <MyAsyncFunctionComponent message={"message"} />;
+
+		function* MyGeneratorComponent(this: Context, props: MyProps) {
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			for (props of this) {
+				let newMessage: string = props.message;
+				// @ts-expect-error
+				let newUnexpected = props.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		}
+		// @ts-expect-error
+		elem = <MyGeneratorComponent />;
+		elem = <MyGeneratorComponent message={"message"} />;
+
+		async function* MyAsyncGeneratorComponent(this: Context, props: MyProps) {
+			let message: string = props.message;
+			// @ts-expect-error
+			let unexpected = props.unexpected;
+
+			for await (props of this) {
+				let newMessage: string = props.message;
+				// @ts-expect-error
+				let newUnexpected = props.unexpected;
+				yield <div></div>;
+			}
+
+			return <div></div>;
+		}
+		// @ts-expect-error
+		elem = <MyAsyncGeneratorComponent />;
+		elem = <MyAsyncGeneratorComponent message={"message"} />;
+	});
+});
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,18 @@ import {isPromiseLike, MaybePromise, MaybePromiseLike, Pledge} from "./pledge";
 // re-exporting EventMap for user extensions
 export {EventMap} from "./events";
 
+declare global {
+	module JSX {
+		interface IntrinsicElements {
+			[tag: string]: any;
+		}
+
+		interface ElementChildrenAttribute {
+			children: {};
+		}
+	}
+}
+
 type NonStringIterable<T> = Iterable<T> & object;
 
 function isIterable(value: any): value is Iterable<any> {
@@ -24,6 +36,7 @@ export type Tag<TProps = any> = Component<TProps> | string | symbol;
 // prettier-ignore
 type TagProps<TTag extends Tag> = 
 	TTag extends Component<infer TProps> ? TProps 
+	: TTag extends string ? JSX.IntrinsicElements[TTag]
 	: any;
 
 export type Key = unknown;
@@ -98,18 +111,6 @@ export type Portal = typeof Portal;
 
 export const Raw = Symbol.for("crank.Raw") as any;
 export type Raw = typeof Raw;
-
-declare global {
-	module JSX {
-		interface IntrinsicElements {
-			[tag: string]: any;
-		}
-
-		interface ElementChildrenAttribute {
-			children: {};
-		}
-	}
-}
 
 export function isElement(value: any): value is Element {
 	return value != null && value[ElementSigil];
--- a/src/index.ts
+++ b/src/index.ts
@@ -539,16 +539,12 @@ abstract class ParentNode<T, TProps extends Props = Props>
 	}
 }
 
-class FragmentNode<T, TProps extends Props> extends ParentNode<T, TProps> {
+class FragmentNode<T> extends ParentNode<T, any> {
 	readonly tag: Fragment = Fragment;
 	readonly key: Key;
-	readonly parent: ParentNode<T, TProps>;
+	readonly parent: ParentNode<T>;
 	readonly renderer: Renderer<T>;
-	constructor(
-		parent: ParentNode<T, TProps>,
-		renderer: Renderer<T>,
-		key: unknown,
-	) {
+	constructor(parent: ParentNode<T>, renderer: Renderer<T>, key: unknown) {
 		super();
 		this.key = key;
 		this.parent = parent;
@@ -578,10 +574,10 @@ class FragmentNode<T, TProps extends Props> extends ParentNode<T, TProps> {
 	}
 }
 
-class HostNode<T, TProps extends Props> extends ParentNode<T, TProps> {
+class HostNode<T> extends ParentNode<T, any> {
 	readonly tag: string | symbol;
 	readonly key: Key;
-	readonly parent: ParentNode<T, TProps> | undefined;
+	readonly parent: ParentNode<T> | undefined;
 	readonly renderer: Renderer<T>;
 	value: T | undefined;
 	private childValues: Array<T | string> = [];
@@ -589,7 +585,7 @@ class HostNode<T, TProps extends Props> extends ParentNode<T, TProps> {
 	private iterator: Iterator<T> | undefined = undefined;
 	private readonly hostCtx: HostContext<T>;
 	constructor(
-		parent: ParentNode<T, TProps> | undefined,
+		parent: ParentNode<T> | undefined,
 		renderer: Renderer<T>,
 		tag: string | symbol,
 		key?: unknown,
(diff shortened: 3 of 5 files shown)


monodop:
I think everything is working now. If you look [here](src/__tests__/types.tsx) you can see all of the typing tests, but they only work in TypeScript 3.9.0, since they rely on the new `// @ts-expect-error` attribute. Before this gets merged we can probably just comment them out until the new typescript version comes out.

@brainkim are you ok with these changes? I made sure it works with looser typing. If you have your tsconfig set to allow implicit this, you can get even looser and I think everything still works.

brainkim:
@monodop I’ll need to pull down the changes and play around with it. I promise to get you concrete feedback by tomorrow night.

> they only work in TypeScript 3.9.0, since they rely on the new // @ts-expect-error attribute

If it’s just for the tests then I don’t mind updating, but I’m typically pretty conservative on adopting new TypeScript features (still haven’t adopted nullish coalescing or optional chaining)

> If you have your tsconfig set to allow implicit this

I like to throw `any` into my code here and there but I always run typescript with `--strict`

monodop:
Ok, to get the tests running you have to run `npm install --save-dev typescript@3.9.0-beta` (or whatever the yarn equivalent is). It might be a good idea to wait to put that into the main package until it's actually released though. It's slated for mid-may currently so that's not too long to have those specific tests disabled.

Roadmap: https://github.com/microsoft/TypeScript/issues/37198

If you're not ok with that, I might be able to get `dtslint` or `tsd` up and running with some hacky workarounds but my initial attempts at that weren't very successful.

src/index.ts
```diff
@@ -42,35 +41,44 @@ export interface IntrinsicProps<T> {
...
 // because this breaks Function.prototype methods.
 // https://github.com/microsoft/TypeScript/issues/33815
-export type Component = (
-	this: Context,
-	props: Props,
-) => ChildIterator | MaybePromiseLike<Child>;
+export type Component<TProps = any> =
+	| GeneratorComponent<TProps>
+	| FunctionComponent<TProps>
+	| (TProps extends {} ? never : MaybePromise<Child>);
```
brainkim, line 81: I’m not sure this is correct. For Crank, all components are functions, and anything else isn’t a component. This has concrete ramifications when it comes to `createElement` calls (`createElement({})` should be a type error.

src/index.ts
```diff
@@ -108,14 +119,14 @@ export function isElement(value: any): value is Element {
 	return value != null && value[ElementSigil];
 }
 
-export function createElement<TTag extends Tag>(
+export function createElement<TTag extends Tag, TProps extends Props>(
```
brainkim, line 122: I think we might need to think more about the way to type createElement calls. The `Tag` type and its props are implicitly connected, insofar as if `Tag` is a function, the props type is the first parameter (maybe accessible through `Parameters<TTag>`.

As for intrinsic elements, the problem is that TypeScript and JSX already offer a means of typing intrinsic elements, which is the `IntrinsicElements` object on the `JSX` interface.

Not using some kind of inference to link the Tag and Props type seems wrong, but I’m not sure how it would work.

src/index.ts
```diff
@@ -954,9 +967,9 @@ export class HostContext<T = any> {
 
 export interface ProvisionMap {}
 
-const componentNodes = new WeakMap<Context, ComponentNode<any>>();
-export class Context extends CrankEventTarget {
-	constructor(host: ComponentNode<any>, parent?: Context) {
+const componentNodes = new WeakMap<Context<any>, ComponentNode<any, any>>();
```
brainkim, line 970: Even for me, `ComponentNode<any, any>` starts to look a bit too wild haha.

src/index.ts
```diff
@@ -936,14 +947,16 @@ function createNode<T>(
...
 		return new HostNode(parent, renderer, child.tag, child.key);
+	} else {
+		throw new Error('Unsupported child/tag type');
 	}
 }
 
-const hostNodes = new WeakMap<HostContext<any>, HostNode<any>>();
-export class HostContext<T = any> {
-	constructor(host: HostNode<T>) {
+const hostNodes = new WeakMap<HostContext<any>, HostNode<any, any>>();
```
brainkim, line 957: Typing props for `HostContexts` is gonna be tricky, might have to think more about this.

src/__tests__/types.tsx
```diff
@@ -0,0 +1,239 @@
...
+import { createElement, Component, FunctionComponent, Context, GeneratorComponent, Child } from "..";
+
+describe("types", () => {
+
+    type MyProps = {
+        message: string;
+    };
+    let elem: any;
+
+    test("Primitive Components", () => {
```
brainkim, line 11: See the comment about the component types. Components are functions, and these values are captured by the `Child` type.

src/index.ts
monodop: `createElement({})` is a type error with this typing. The `(TProps extends {} ? never : MaybePromise<Child>)` simply means that if you specify TProps to be anything that extends `{}`, then `Component<TProps>` will no longer include `MaybePromise<Child>`. The reason for this is to say that `Component<any>` matches `Child` but `Component<MyProps>` doesn't. My understanding is that `Child` should be allowed in `Component<any>` but I'm not 100% sure if that's correct.

src/index.ts
monodop: What are the drawbacks of leaving it as is? If this produces the correct types in the api for the user, will it cause any issues?

src/index.ts
monodop: I haven't fully grasped how `HostContext` works, this is mostly a guess. Is `any` not sufficient?

src/__tests__/types.tsx
monodop: Ok, this is something I hadn't exactly considered. I can remove `MaybePromise<Child>` from the Component definition and see how that affects things

src/index.ts
brainkim: Ah sorry. The types can be confusing because they’re circularly defined: 
`Components` are functions which can yield or return `Elements`. `Elements` have `tags` which are strings, symbols or `Components`. The following `createElement` calls would also errors:
```js
createElement(true);
createElement(0);
createElement([]);
createElement(null);
```

I think maybe you were thinking of `Child` or `Children` types, which are huge types which represent parts of the element tree.

monodop: Sounds good! I pushed the fix for it

src/index.ts
monodop: Any ideas? do you want me to make `any` default?

src/index.ts
```diff
@@ -108,14 +119,14 @@ export function isElement(value: any): value is Element {
...
 
-export function createElement<TTag extends Tag>(
+export function createElement<TTag extends Tag, TProps extends Props>(
 	tag: TTag,
-	props?: Props | null,
+	props?: TProps | null,
 	...children: Array<Children>
 ): Element<TTag>;
-export function createElement<TTag extends Tag>(
+export function createElement<TTag extends Tag, TProps extends Props>(
```
codejet, line 127: why does `createElement` actually exist twice @brainkim (and the first declaration not having a body)?

brainkim: @codejet this is TypeScript’s function overloading syntax.

codejet: thanks @brainkim. i'm still relatively new to ts.

src/index.ts
brainkim: Sorry it still doesn’t seem to be correct? Like Component should just be union of FunctionComponent/GeneratorComponent but there’s still `(TProps extends {} ? never : MaybePromise<Child>)`? Maybe I’m looking at a stale branch.

src/index.ts
brainkim: I’m pretty conservative about adding type parameters because each type parameter is a possible breaking change for TypeScript users if we decide to change or remove it at a later point. And if there’s a way to get createElement calls to infer the correct props from the `tag` argument that would be the best of all possible worlds. The type info exists on the [JSX IntrinsicElements interface](https://www.typescriptlang.org/docs/handbook/jsx.html#intrinsic-elements) and the component’s first parameter; the tricky thing is figuring out how to pull it in.

src/index.ts
monodop: My bad, I messed up my commit. It should be there for real this time :)

src/index.ts
monodop: I'll see if I can get type inferencing to work between the two arguments, but it seems like if you use tsx, the compiler/linter already can figure out the appropriate types from the tag you've provided without the inference.

src/index.ts
monodop: Actually this is a good question. Why do we have an overload with `children` if we never use it in `createElement`? The actual function definition doesn't reference it.

src/index.ts
monodop: New `createElement` signature, seems to infer properly:

```tsx
export type Tag<TProps=any> = Component<TProps> | string | symbol;
type TagProps<TTag extends Tag> = TTag extends Component<infer TProps> ? TProps : any;

export function createElement<TTag extends Tag>(
	tag: TTag,
	props?: TagProps<TTag> | null,
	...children: Array<Children>
): Element<TTag>;
```

image

image

image

monodop: I'm wondering if we should fallback to `any` for string/symbol components. Maybe instead it should fallback to `{[prop: string]: any}`? Or maybe we can type certain common html elements?

monodop: If you wanted to fallback on the `JSX.IntrinsicElements` type you described above, I can update the TagProps type to look like the following, but it still has the same sort of issues, considering we're still defining each tag as any anyway:

```tsx
type TagProps<TTag extends Tag> = 
	TTag extends Component<infer TProps> ? TProps : 
	TTag extends string ? JSXIntrinsicElements[TTag]:
	any;
```

brainkim: Holy shot you did it! Sorry, I was in the hole all day debugging some issues. I’ll look at this ASAP.

monodop: No worries! It's been a long weekend for me with some random issues too so I get it :)

Just let me know if this works for you or if there's more changes. Can also look into typing the intrinsic elements too if you have some ideas, just it might be somewhat of an undertaking if you want it all done at once.

brainkim: > It's been a long weekend for me

It’s Wednesday 😅 Hoping you’re okay in these difficult times.

src/index.ts
```diff
@@ -45,32 +49,34 @@ const ElementSigil: unique symbol = Symbol.for("crank.ElementSigil");
 export interface Element<TTag extends Tag = Tag> {
 	[ElementSigil]: true;
 	readonly tag: TTag;
-	props: Props;
+	props: TagProps<TTag>;
```
brainkim, line 52: This is cool.

src/index.ts
```diff
@@ -691,7 +699,7 @@ class ComponentNode<T> extends ParentNode<T> {
 			return [undefined, undefined];
 		} else if (this.iterator === undefined) {
 			this.ctx.clearEventListeners();
-			const value = new Pledge(() => this.tag.call(this.ctx, this.props!))
+			const value = new Pledge(() => this.tag.call<Context<TProps>, [NonNullable<TProps>], MaybePromise<Child> | ChildIterator>(this.ctx, this.props!))
```
brainkim, line 702: I was wondering how you managed to invoke `Function.prototype.call` on the union of `GeneratorComponent` and `FunctionComponent`, but I prefer the old way of typing `Component`?

See https://github.com/microsoft/TypeScript/issues/33815 for why I chose to use a union in the return value versus a union of functions. It‘s just really wonky in TypeScript right now.

src/index.ts
```diff
@@ -45,32 +49,34 @@ const ElementSigil: unique symbol = Symbol.for("crank.ElementSigil");
...
-export type GeneratorComponent = (this: Context, props: Props) => ChildIterator;
+export type GeneratorComponent<TProps = any> = (
+	this: Context<TProps>,
+	props: TProps
+) => ChildIterator;
 
 // TODO: component cannot be a union of FunctionComponent | GeneratorComponent
 // because this breaks Function.prototype methods.
 // https://github.com/microsoft/TypeScript/issues/33815
-export type Component = (
```
brainkim, line 70: I prefer the old union rather than a union of function types. See https://github.com/bikeshaving/crank/pull/51/files#r417747284 and https://github.com/microsoft/TypeScript/issues/33815 for more info.

src/__tests__/types.tsx
```diff
@@ -0,0 +1,249 @@
...
+        // @ts-expect-error
+        const MyNull: Component = null;
+
+        // @ts-expect-error
+        const MyUndefined: Component = undefined;
+
+        // Not entirely sure why this one doesn't error anymore. seems that <a></a> returns
+        // any, while createElement('a') returns Element<any>.
+        // skip @ts-expect-error
+        // const MyElement: Component = <a></a>;
```
brainkim, line 31: We can specify the type of a JSX element by defining the JSX.Element on the global JSX type. I think. I’m not 100% sure but that’s what the docs say. https://www.typescriptlang.org/docs/handbook/jsx.html#the-jsx-result-type

src/__tests__/types.tsx
```diff
@@ -0,0 +1,249 @@
+/** @jsx createElement */
```
brainkim, line 1: I really really love these type tests. The one thing I’m worried about is that the jest linter will possibly complain about it because none of these tests have actual assertions. Also we should probably just upgrade to TypeScript 3.9 RC in package.json if we’re going to be using these features (https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-rc/)

src/index.ts
monodop: I did see that, but I was able to get typescript to be happy. I'd prefer component to be both, since it _can_ be both, right?

monodop:
> Love the name of the commit “Solved the mystery”😄

Yeah I felt stupid after I figured out what I was doing wrong. I'll squash most of these commits once everything's settled down a bit so the history is a little cleaner.

src/__tests__/index.tsx
```diff
@@ -343,9 +343,12 @@ describe("sync generator component", () => {
 	});
```
brainkim, line 343: One thing with the tests that aren’t types.tsx, I’ve recently did a little refactoring to these files and we might encounter some nasty rebases. Also I think a cool thing would be if we could use the loosely typed versions of typings in the original tests (we can always change them later but it would be proof that the new typings are backwards compatible). In other words, let’s see if we can get away with not changing any of the test files for the new typings.

src/__tests__/types.tsx
brainkim: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/expect-expect.md this is the rule that I expect to start throwing. Maybe we can disable it for the file.

src/index.ts
```diff
@@ -92,15 +98,17 @@ export type Portal = typeof Portal;
 export const Raw = Symbol.for("crank.Raw") as any;
 export type Raw = typeof Raw;
 
+interface JSXIntrinsicElements {
```
brainkim, line 101: Curious about your reasoning for moving the interfaces outside the global. Also I’ve been looking at Preact’s TypeScript typings and it looks like they define JSX as a non-global module? This would be super interesting to look into but maybe can be done in a separate PR. I’m not sure how global module stuff actually works in TypeScript.
https://github.com/preactjs/preact/blob/master/src/index.d.ts#L7

src/index.ts
brainkim: The thing is, generator components are still functions it’s just the return type that’s different. The reason I would want to use the old type is suddenly the union function type can’t be used with call and apply, and at the end of the TypeScript is all about structural typing anyways, so redeclaring types which theoretically have the same structure isn’t a big deal.

monodop: In my opinion the types should represent what things are. A component can be a function that returns a child or a promise of a child, or it can be a generator or async generator that yield childs. I get that you can't use .call or .apply on it right now, but that's not really used very many places in the codebase. Are users going to be using .call or .apply? Can we just use hacky workarounds in our own code for now until typescript eventually decides to do something so that it's a nicer experience for the end user?

src/__tests__/types.tsx
monodop: I'll check it out

src/__tests__/types.tsx
monodop: I haven't seen any linter errors when I run `npm run test`, but maybe I'm missing something. I'll update package.json with the rc version - was just hesitant to do so unless you thought it was a good idea. I don't know if there's any implications for users if we do this though.

src/__tests__/index.tsx
monodop: Ok, I don't think anything will break but I'll see if I can get the original tests to work.

src/index.ts
monodop: I moved it out because I wasn't able to reference `JSX.IntrinsicElement` for some reason. I'm not sure why but I can look into some other options if it bugs you :)

src/__tests__/types.tsx
brainkim: We’re using yarn so it would be `yarn run test`. But anyways the linting stuff is `yarn run lint` and you can fix prettier errors automatically with `yarn run lint --fix`.

brainkim: We definitely need a CONTRIBUTING.MD

monodop:
I have no issues with rebasing and rewriting commits for this. I do it all the time at my job and I think it really keeps things looking nice. Usually I try to preserve smaller logical commits if there's atomic changes that make up the full PR. I just don't do it until the end because it makes it harder to review if everything's being rebased all the time. I think you'll be happy with it :)

brainkim:
If you’re happy to rebase and squash I’m happy. But know that I will gladly help if the rebase turns out to be hairy (we both touched index.ts a little bit).

src/index.ts
brainkim: Hmmm we can change it later cuz it isn’t exported directly.

src/__tests__/types.tsx
monodop: Ah ok. I've actually never used yarn before so I'll look into it. Would be nice to get that all automated in github at some point 😄

src/index.ts
brainkim: I mean... the issue was opened last year in October and has yet to receive a comment from any of the TypeScript maintainers. And yes, I do want to leave open the possibility for advanced users to use `call` or `apply` on a component type even if they shouldn’t. Who knows maybe there’s an advanced use-case for routers or something.

At the end of the day, `FunctionComponent | GeneratorComponent` is structurally the same as `(this: Context, props: TProps) => ChildIterator | MaybePromiseLike<Child>` and I’d prefer the solution which caused fewer hacks, given that TypeScript is structurally typed.

src/index.ts
brainkim: Ehhh we can fix it later if necessary.

src/index.ts
```diff
@@ -1045,10 +1055,10 @@ export class Renderer<T> {
 		if (isElement(child) && child.tag === Portal) {
 			portal = child;
 		} else {
-			portal = createElement(Portal, {root}, child);
+			portal = createElement<Portal>(Portal, {root}, child);
```
brainkim, line 1058: Shouldn’t this be inferred with the new typings?

src/index.ts
monodop: You know what? I just realized I was arguing about the wrong thing. With all the changes I was doing while testing stuff, I forgot what the original typing was and I think they're both pretty much the same thing. For some reason I thought the ChildIterator part wasn't in the original typing. So yeah I think you're right and I'll change it back.

src/index.ts
monodop: Probably, except Portal is typed as `any` for some reason, so createElement gets typed as `any` too. Not sure if that matters though.

image

src/index.ts
monodop: Done

src/index.ts
monodop: Done

src/__tests__/types.tsx
monodop: Done

src/__tests__/types.tsx
monodop: I spent about 15 minutes playing with this, and it seems super complicated. I think `JSX.Element` needs to be typed as `MaybePromise<Child>`, and anywhere you used `Element` in the tests should really be `MaybePromise<Child>` or `Child`. `renderer.render` will need to accept `MaybePromise<Child>` too. 

I think the types on `ChildIterator` are wrong, should probably be `Iterator<Child, Child | void, TNext>` and `AsyncIterator<Child, Child | void, TNext>`.

It gets really ugly because you have to use `MaybePromise<Child>` a lot in the tests. I tried wrapping `Child` with `MaybePromise` so that it's always included, but then async components break because they are not `MaybePromise<Child>`, they're actually `Promise<Child>`. 

This change will definitely break a lot of existing code that rely on `Element`. Might be worth making a separate PR for it.

brainkim: Doesn’t have to be part of the PR if you’re eager to get these changes shipped.

src/__tests__/index.tsx
monodop: Done!

monodop:
Ok I think this might be ready to go. I noticed I updated some of the types from `MaybePromiseLike` to `MaybePromise`. Do you think this will cause any issues? I'm not too sure what the practical difference is.

brainkim:
@monodop I prefer to have components return promise-likes rather than promises. Promise-likes means something which implements the `then` method as specified by the promises/A+ spec. Promises are the actual ES6 promise class, which includes the methods `catch` and `finally`. Even if most TypeScript people are using regular promises, I want Crank to follow [the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) and accept promise-likes even if we work output promises exclusively.

monodop:
Okay, I put the MaybePromiseLike's back. Anything else?

brainkim:
@monodop I need to think a little more about the way props work as they flow into the hidden `ParentNode` classes. For instance, you set the ParentNode props to the Props of the current node, which I think is incorrect. It implies that every node of the element tree has the same props shape, which isn’t true. I also need to think a little about HostNode `props`, and how Props relate to `IntrinsicProps`. Finally, I’m also considering dropping the requirement of having `children` be of the `Children` type. I’ve seen Crank code in the wild which uses render props for children, and while I hate the pattern, I don‘t want to prevent it with TypeScript.

I’m trying not to be a finicky code reviewer, sorry 😖.

monodop:
No worries, I'm fine with iterating on this until it's right. I felt a little weird typing out all the node elements, but it at least it all compiles at this point. I found the naming for a lot of these things to be kind of confusing so I wasn't exactly sure how the types needed to be set. I'm not sure what all the node types are supposed to be, since they all extend ParentNode, but really they should just be regular Nodes, right?

monodop:
What do you think about the changes in https://github.com/bikeshaving/crank/pull/51/commits/b6ba7144a82091dda46ccab1b66e41a60e210a4f

package.json
```diff
@@ -42,6 +42,8 @@
     "shx": "^0.3.2",
     "ts-jest": "^25.3.1",
     "typescript": "3.9.1-rc"
+    "husky": "^3.1.0",
```
brainkim, line 45: Oof I beansed it.

brainkim: How do I fix it

brainkim: I have it correct on my local but am unable to push to this branch for some reason 😓

brainkim:
Closed for https://github.com/bikeshaving/crank/pull/96/
Most work was preserved, and just added a single commit which tweaks some stuff:
https://github.com/bikeshaving/crank/pull/96/commits/22cac8a3ff717e051b2710e7d85e56258831da38

Thanks for working on this stuff @monodop!

monodop:
Oh hey sorry I didn't get back to you on this one. Thanks for cleaning it up and getting it in!

brainkim:
@monodop Sorry I messed up the merge! And again, super impressed with the TS-foo. `TagProps` seems to work perfectly!

monodop:
Why did you untype context from ComponentNode?

brainkim:
@monodop line number?

monodop:
https://github.com/bikeshaving/crank/pull/96/commits/22cac8a3ff717e051b2710e7d85e56258831da38?file-filters%5B%5D=.ts#diff-f41e9d04a45c83f3b6f6e630f10117feL689

brainkim:
Ahhhh hmmm I think I was removing TProps stuff from parent node and I had a type error with ComponentNode and inheritance, but it probably could be more strictly typed there. Good catch, my bad.

monodop:
I also thought it was odd to untype Props from tag as well from ComponentNode.

monodop:
If there's anything I can do to help fix any weird type errors let me know and I can take a stab at it. Maybe needs a new ticket or something though 🙂

brainkim:
> I also thought it was odd to untype Props from tag as well from ComponentNode

If you’re asking why we don’t use the `Props` type from being used it’s mostly because I didn’t want to prevent render props as children via types, which is what that type would do if someone writes a render prop and then iterates over `this`.

brainkim:
@monodop Definitely! I’m always interested in better typings and if you ever want to get a deeper understanding of the admittedly intense async code which I haven’t commented yet, I’m down to give a walkthrough. Might do a twitch or discord stream at some point with people interested to lower bus factor of the code.

monodop:
> > I also thought it was odd to untype Props from tag as well from ComponentNode
> 
> If you’re asking why we don’t use the `Props` type from being used it’s mostly because I didn’t want to prevent render props as children via types, which is what that type would do if someone writes a render prop and then iterates over `this`.

Nope, was more asking about why you changed `tag` from `Tag<TProps>` to `Tag` (effectively `Tag<any>`)

monodop:
> @monodop Definitely! I’m always interested in better typings and if you ever want to get a deeper understanding of the admittedly intense async code which I haven’t commented yet, I’m down to give a walkthrough. Might do a twitch or discord stream at some point with people interested to lower bus factor of the code.

I'd be super interested depending on the time difference. Is there already a discord or some other chat we can use for this project? This in particular is getting a little off topic but I'd like to continue the discussion

brainkim:
@monodop Ahhh if you’re asking about the tag value in ParentNode, then it’s because I removed the TProps type parameter from the ParentNode class. I do want to type it at some point it just gives me a headache to think about right now how Tag<TProps> would work with inheritance, and I’m not really sure about the value of typing it on the internal nodes yet. What was important for the PR was that Component and Context were properly typed for dependents and we can figure out the nitty gritty about how TProps flow through the internal node classes in another PR I guess.

monodop:
> @monodop Ahhh if you’re asking about the tag value in ParentNode, then it’s because I removed the TProps type parameter from the ParentNode class. I do want to type it at some point it just gives me a headache to think about right now how the TProps parameter would work with inheritance, and I’m not really sure about the value of typing it on the internal nodes yet. What was important for the PR was that Component and Context were properly typed for dependents and we can figure out the nitty gritty about how TProps flow through the internal node classes in another PR I guess.

Ok, makes sense. Having type safety on the component signature was my biggest goal here, and I think we've got that nailed down. Was just curious about the details 👍

brainkim:
No discord yet. Need to work on community stuff but right now I’m just trying to build on the momentum of the initial release. Lots of work to do and happy for all the help! I’m on discord at brain#4221 if you want to contact me directly though!

ryhinchey:
happy to help with community stuff eventually

Issue #33, opened 2020-04-18
Getting refs


dyst5422:
Seems like getting a ref as shown in the documentation doesn't work on the first render
![Screen Shot 2020-04-18 at 10 02 38 AM](https://user-images.githubusercontent.com/13282627/79644128-c11afd80-815b-11ea-899a-29c83a3094bf.png)

results in 
![Screen Shot 2020-04-18 at 10 03 16 AM](https://user-images.githubusercontent.com/13282627/79644154-e60f7080-815b-11ea-80f9-40c6354ab0b9.png)
on the first render and
![Screen Shot 2020-04-18 at 10 03 27 AM](https://user-images.githubusercontent.com/13282627/79644156-ec055180-815b-11ea-9fd8-01ee4f1e2929.png)
after button click

brainkim:
Yeah this is my bad! I wrote that ref code real quick without thinking about it and what you’re pointing out highlights the difficulty of using sync generators with refs. Your code would work if you converted it to an async generator but sometimes you want things to happen synchronously and I gotta respect that. The problem is that synchronous generators suspend at their yield, which means that if you assign the yielded expression to something, it will actually have paused *right* before the assignment, which is kinda frustrating. And the ref will only be the value previously rendered which sucks.

The solution is either to use an async generator, which continuously executes while the component is mounted, or do something like this:

```jsx
function *MyComponent() {
  let ref;
  Promise.resolve().then(() => this.refresh());
  ref = yield <div />
  while (true) {
    ref.focus();
    yield <Copy />;
  }
}
```

This way the component refreshes once and has the ref variable set to the yield. This is sorta emulating the componentDidMount workflows for React.

I need to update the docs to get them to work but this is basically how to do it with sync generators. The great thing is that Crank doesn’t care if you perform side-effects in the course of rendering so it doesn’t really matter if you’re doing stuff before update or after update.

dyst5422:
This is a real bummer as I'm sure it will be an issue many run into. I imagine this should be highlighted in the FAQ.

brainkim:
Believe me I tried really hard to make refs work in sync generators just like they do in async generators, but I couldn’t figure it out. I thought about stepping through generators twice per update, to make sure that refs were assigned, but this made the execution of generators more opaque and didn’t even solve the problem. I thought maybe we could have a callback API for getting the ref but this seemed unwieldy.

[...]

Issue #10, opened 2020-04-16
addEventListener and delegation with tagName is awkward


tomhicks:
Not an issue per se (like a bug or anything), but having just read through all the blog post and docs, this was the one bit that I thought "modern React" definitely does better at.

No idea if this is even possible, but I'm sure I won't be the only one turned off by this.

brainkim:
Thank you for the feedback and congrats on being the first issue in this repository. I anticipated this concern (why use `this.addEventListener("click", callback)` rather than `onClick={callback}`?) and have a coked out journal entry describing reasons for this design decision which I will try to distill in this issue.

I didn’t want to follow the `onClick` callback API because over the years I‘ve found it lacking in several key respects:

1. `onClick`-APIs are a leaky abstraction.
EventTargets in the DOM take more than just a function; they take several options including `once`, `capture` and `passive`. React has attempted to model capture with `onClickCapture`-type props, and `once` event listeners can be emulated with state, but there is still currently [no solution for registering passive event listeners](https://github.com/facebook/react/issues/6436), and you can sorta see the folly of using props, which are singular values, to model EventTarget listeners (`onHoverCapturePassive`?) I guess you could make the `onClick`-type props take a tuple, but that seems immediately more unwieldy that the API Crank proposes. In fact, there has been some work trying to support more of the [EventTarget API](https://github.com/facebook/react/pull/17651) with hooks to address these concerns.

2. `onClick`-APIs make JSX look terrible.
I know that you can assign callbacks to a variable and pass them into JSX by reference, but so long as it’s possible to write event listeners inline, people will write them inline. And when you do, you lose a lot of the similarities JSX has with HTML, because functions are multiline syntax structures and don’t map to anything you can do in HTML. IMO, inline functions in JSX are just plain ugly.

3. `onClick`-APIs force you to dox anonymous event listeners.
Even when you define your callbacks as variables before passing it into JSX, it can be somewhat taxing to have to come up with names for these callbacks. When I’ve used React, I would typically use conventions like `this.handleClick` or `const handleModalClick =` for callbacks which would be passed to child components, but I still felt untethered. Using the `this.addEventListener` API removes this specific pain point, by allowing callbacks to be understood by usage rather than named.

4. Mimicking `onClick`-APIs is a pain when writing components.
[...]

PR #52, opened 2020-04-24
Updates to docs

brainkim:


brainkim:
Addresses some concerns in #10 and #33. More work on the refs stuff is needed but requires some changes to API so I’m going to merge for now.

PR #63, opened 2020-04-25
Fixes

brainkim:
Some bug fixes that didn’t make launch:
1. Wrapping non-string iterables returned from components in a fragment.
Previously Crank had the behavior where a component which returned or yielded an iterable would treat that iterable as its children. This led to surprising behavior where yielding a `Copy` after yielding an iterable would result in only the first element of the iterable being preserved. This change makes sure that the iterable is treated as a single unit, which is less surprising.
2. Unmounting stateless render roots.
Previously stateless server renderings didn’t unmount generator components. This can lead to zombie processes in node. We now unmount components after every render when root object isn’t provided.


brainkim:
Deployed in v0.1.1

Issue #59, opened 2020-04-24
Github Actions, Travis, Circle ?


ryhinchey:
@brainkim do you have a preference for one of these services?  With more PRs coming in, it'd be great to have a CI service run tests, the linter and a build.

Happy to set one of these up!

brainkim:
Probably GitHub Actions is the way to go! Haven’t had a chance to play with them and will see what can be done.

PR #76, opened 2020-04-28
adding initial github actions config

ryhinchey:
This PR adds a basic github actions workflow to Crank.  It'll run for node 10, 12, and 14.  Runs the build, tests, and linting.

Here's an example of the output:
https://github.com/ryhinchey/crank/pull/1/checks?check_run_id=624688094

`--frozen-lockfile` is needed in the install step so yarn won't install optional dependencies like `fsevents` that will break the build

Closes https://github.com/bikeshaving/crank/issues/59

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,24 @@
+name: Crank CI Workflow
+
+on:  
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


brainkim:
This is kinda cool goddamn GitHub. Thanks for doing this! Will merge ASAP. Also the tests could probably be a lot faster I just am afraid of using fake timers for the microtask stuff because this has bitten me in the butt before.

ryhinchey:
Everything can be optimized in time! Nice work on the tests so far

brainkim:
GitHub actions, pretty amazing.

Issue #82, opened 2020-04-29
Reusable component logic with setup and teardown


wmadden:
One really common problem with React's imperative style of class components was that it's very difficult to write code that's reused between components and involves both setup and teardown steps without using higher order components.

I can't think of an easy way to achieve this in plain JavaScript generators. How would you go about solving this problem in Crank?

If there's no immediate solution, it might be a good goal to keep in mind as the component API is evolving.

monodop:
I think after thinking about this, one of the hardest things will be dealing with sync vs async code. Consider that you want to implement some form of abstraction (let's just call it a plugin), you'd start with four types of situations that need to be handled:

Component is sync, Plugin is sync
Component is sync, Plugin is async
Component is async, Plugin is sync
Component is async, Plugin is async

On top of that, you may decide to allow plugins to call other plugins, which could have similar types of possibilities.

What this means is: What if my plugin setup logic needs to happen async, but I want to use it in a sync component? or visa versa? What if I want my setup to run async but provide some kind of loading result so that the component can run synchronously or provide a loading spinner or something? What if I need to respond to changes to props async before the component gets them? Or do something after the component rendered? And does the order of these actions matter when there's multiple plugins registered at once? And so on.

We should probably be really careful in designing the api so that we don't create situations where a user has to make concessions about their components to fit whatever third party plugin they're using. But at the same time we need the user to have some kind of control over what's going on without the whole thing being really confusing and having plugins work different ways depending on if you need sync or async. Ideally it should Just Work™. I'm not sure if we'll ever hit that goal if we need to support both sync and async components out of the box, but it is an interesting problem to think about.

ryhinchey:
@wmadden what do you mean by setup and teardown?  Are you thinking of `componentDidMount` and `componentWillUnmount`?

As @monodop pointed out, there are many permutations of component and plugin types - it gets complicated quickly.

Perhaps we could leverage Crank's event system?  This idea isn't fully thought through but it does show how the event system can be used to communicate between both sync/async generators and "utility code" that's used in several components.  

https://codesandbox.io/s/wispy-surf-qbcki?file=/src/index.js

```jsx
/** @jsx createElement */
import { createElement } from "@bikeshaving/crank";
import { renderer } from "@bikeshaving/crank/dom";

const LoadingIndicator = () => <h1>loading</h1>;

// fake user api
const getUser = () =>
  new Promise(resolve => {
    setTimeout(() => {
      resolve({ name: "Jenny" });
[...]

PR #86, opened 2020-05-01
add unmount event

ryhinchey:
This PR adds an `unmount` event that gets fired when a component unmounts. This allows you to do something like this:

```jsx
function App() { 
  const id = setInterval(() => {
    console.log('something');
  }, 1000);

  this.addEventListener('unmount', e => {
    clearInterval(id);
  });

  return (
    <h1>...</h1>
  )
}
```

The event does not bubble up to parents.

This is the first event fired from within Crank core out to components.  It's been mentioned a few times that having an `unmount` event would be handy as an alternative to `try/finally` and I think this would be a nice addition to Crank.  It also opens the door to discuss what other events Crank might want to fire.  Do we want to namespace these events (something like `crank:unmount`)?

If we're on board bringing this change in, I'll update the docs.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -865,6 +865,7 @@ class ComponentNode<T> extends ParentNode<T> {
 		}
 
 		this.updating = false;
+		this.ctx.dispatchEvent(new CustomEvent("unmount"));
 		this.unmounted = true;
 		this.ctx.clearEventListeners();
 		if (!this.finished) {
--- a/src/__tests__/index.tsx
+++ b/src/__tests__/index.tsx
@@ -1764,3 +1764,75 @@ describe("parent-child refresh cascades", () => {
 		});
 	});
 });
+
+describe("events fired within Crank", () => {
+	test("unmount event will fire", () => {
+		const onUnmount = jest.fn();
+
+		function Name(this: Context) {
+			this.addEventListener("unmount", onUnmount);
+			return <h1>crank</h1>;
+		}
+
+		function* Component(this: Context): Generator<Element> {
+			let visible = true;
+
+			const onclick = () => {
+				visible = !visible;
+				this.refresh();
+			};
+
+			for (const _ of this) {
+				yield (
+					<div>
+						<button onclick={onclick}>Toggle Name</button>
+						{visible && <Name />}
+					</div>
+				);
+			}
+		}
+
+		renderer.render(<Component />, document.body);
+
+		const button = document.querySelector("button")!;
+		button.click();
+
+		expect(onUnmount).toHaveBeenCalledTimes(1);
+	});
+
+	test("unmount event does not bubble", () => {
+		const onUnmount = jest.fn();
+		const onUnmountInParent = jest.fn();
+
+		function Name(this: Context) {
+			this.addEventListener("unmount", onUnmount);
+			return <h1>crank</h1>;
+		}
+
+		function* Component(this: Context): Generator<Element> {
+			let visible = true;
+			this.addEventListener("unmount", onUnmountInParent);
+			const onclick = () => {
+				visible = !visible;
+				this.refresh();
+			};
+
+			for (const _ of this) {
+				yield (
+					<div>
+						<button onclick={onclick}>Toggle Name</button>
+						{visible && <Name />}
+					</div>
+				);
+			}
+		}
+
+		renderer.render(<Component />, document.body);
+
+		const button = document.querySelector("button")!;
+		button.click();
+
+		expect(onUnmount).toHaveBeenCalledTimes(1);
+		expect(onUnmountInParent).not.toHaveBeenCalled();
+	});
+});


ryhinchey:
@brainkim it looks like I have merge rights for this PR and others.  Might want to look into that 😉

src/index.ts
```diff
@@ -865,6 +865,7 @@ class ComponentNode<T> extends ParentNode<T> {
 		}
 
 		this.updating = false;
+		this.ctx.dispatchEvent(new CustomEvent("unmount"));
```
jspears, line 868: As a random stranger new to crank, trying to be helpful :)

Perhaps, it should include the context of the element that is unmounting, in the event, and also a typed event.  Both could of course be done later...  
```tsx
this.ctx.dispatchEvent(CrankUnmountEvent(this));

```
In react land I'm always trying to hoist functions out, and often can't because enough context isn't given in events to do that.  Why hoist?  To prevent re-renders on propchange mostly, but also increase re-use.

lukejagodzinski:
> Do we want to namespace these events (something like `crank:unmount`)?

I think adding namespace is a good idea. However, in many examples I've seen something like `crank.unmount` and you're proposing `crank:unmount`. Would be good to pick naming convention that most people like. I'm not for any of these conventions. Are there any popular libraries that picked some naming convention?

brainkim:
Thanks for being proactive with this! I know people are sorta clamoring for this kind of functionality. I should have expected that early adopters would be keen on the idea of plugins/mixins.

Early on in the iteration of Crank, I had special Crank events for things like refresh and unmount (https://github.com/bikeshaving/crank/commit/7901d577acfb7ff6ca4f64689293919c86a1e960#diff-116f234c767e9ab5648ef6b4ecfd0bceL906-L908). I removed them after doing some research into event targets and thinking about what using an event target for internal Crank stuff would imply:

1. **Events which do not bubble can still be captured.** Event capturing doesn’t respect the `bubbles` flag, and while we don’t implement event capturing at the component level, we do plan on it, and doing so would be a nightmare if a user chose to listen to a crank unmount event across a large subtree.
1. **Event targets imply a cardinality of zero to many.** There are some events like "close" and "unload" in the DOM which mostly only fire once, but the event target interface in general implies that the callback which is passed to `addEventListener` can be called multiple times. Whatever API we come up with for this stuff should be one and done. The callback is called once and thrown away.
1. **Event targets imply the event can be cancelled with `ev.preventDefault`.** Unmounting is final, and trying to figure out the semantics of what cancelling an unmount event gives me a headache. Does that mean that the unmount itself is prevented? Simply ignoring the result of dispatchEvent doesn’t make much sense.

I think there are too many disanalogies with regard to event targets to use them for the low-level Crank events. I apologize for not being transparent about the way I wanted this API design to go. As an alternative I was thinking about adding two methods to context which simply take a callback: `schedule` and `teardown`. The names can be bikeshed but I like these two because they’re the same number of letters.

`this.schedule` would fire synchronously when the internal node related to a context committed. I like the name `schedule`, because it means we can write code like `this.schedule(() => this.refresh())` which pretty much does what it says on the tin. This would solve the problems mentioned in https://github.com/bikeshaving/crank/issues/33, and it would allow us to enqueue multiple refreshes to a sync generator component synchronously.

`this.teardown` or maybe `this.cleanup` would fire synchronously when the internal node related to a context is unmounted. This would allow people to write extensions to contexts for common globals like `fetch`, `setTimeout` and `setInterval` which would automatically clean themselves up:

```js
const timeoutsMap = new WeakMap<Context, Set<ReturnType<typeof setTimeout>>>();
Context.prototype.setTimeout = function(fn, delay) {
  let timeouts = timeoutsMap.get(this);
  if (timeouts === undefined) {
    timeouts = new Set();
    timeoutsMap.set(this, timeouts);
    this.teardown(() => {
      for (const timeout of timeouts) {
        clearTimeout(timeout); 
      }
    });
  }
  const timeout = setTimeout(fn, delay);
  return timeout;
};
```

This is just a sketch of what writing setTimeout with the teardown callback would look like. We also would need some additional code to handle additional args to setTimeout and maybe clearing timeouts from the set when the timeout actually fires.


Commits and unmounts are the only two internal events I feel comfortable exposing in this way. My intuition is that other events like initialization or refresh could be abused to do terrible things and we do not want to write abstractions which are too powerful/circumvent the need for other patterns like higher-order components. The callbacks should fire synchronously, and probably should be passed the internal childValues of the node they were fired on.

@ryhinchey If you’re interested in implementing let me know and I’d gladly delegate the concrete details of the design to you. Who knows, we may even be able to move EventTarget code to use the schedule/teardown logic rather than being defined as a superclass of Context.

brainkim:
@lukejagodzinski I’m starting to think no separators all lowercase is the best approach.
As in `tododestroy`. Might be somewhat harder to read but matches the way the DOM does it, and we could then do stuff like fire `ontododestory` which have been defined on the props automatically as [suggested here](https://github.com/bikeshaving/crank/issues/10#issuecomment-621574860).

lukejagodzinski:
> @lukejagodzinski I’m starting to think no separators all lowercase is the best approach.
> As in `tododestroy`. Might be somewhat harder to read but matches the way the DOM does it, and we could then do stuff like fire `ontododestory` which have been defined on the props automatically as [suggested here](https://github.com/bikeshaving/crank/issues/10#issuecomment-621574860).

Yep makes sense, I like it.
However, it's a bummer that many event names are not written in camelCase style in DOM where the event name actually is case sensitive. Event names in DOM are actually an example of how inconsistent web is. For example, there is the `DOMContentLoaded` event which has to be written with exactly this casing. The reason behind all other event names being lower case is probably because corresponding attribute names are case insensitive, so it doesn't matter if you write: `<div onclick=""></div>` or `<div onClick=""></div>`, the same is true for `div.setAttribute("onClick", callback)` and `div.setAttribute("onclick", callback)`.
Saying this, I actually wonder if Crank should also be case insensitive if it goes about attribute names. So every attribute would have to be converted to a lower case form. The only problem with this is that it might create confusion and also is might be hard to add TS type to.

ryhinchey:
> I think there are too many disanalogies with regard to event targets to use them for the low-level Crank events. I apologize for not being transparent about the way I wanted this API design to go.

@brainkim thanks for the writeup and general direction!  I agree with all of your points as to why EventTarget isn't an ideal api to used here.  I'll admit it felt strange using it for this but since it was a one line change, I figured why not put it out there.

Two questions for you:

1. How would you describe the difference between a `commit` and `initialization`?
2. We have to account for multiple callbacks registered with `this.teardown` and `this.commit` (naming TBD) right?  `this.teardown` and `this.commit` can only accept one callback but they can  be called multiple times (each call adding a callback). If a context extension uses `this.teardown` and a component uses `this.teardown`, we need to fire both callbacks synchronously on teardown.  Just want to make sure I'm reading that right.

ryhinchey:
related issue https://github.com/bikeshaving/crank/issues/82

brainkim:
> How would you describe the difference between a commit and initialization?

Initialization would be the `constructor` or `componentDidMount` from React. I don’t think we should have any sort of event fire in this instance, and if people want initialization logic they can write it directly at the top of their components. A component `commit` happens when its changes are actually flushed to the DOM. This happens in a parent child relationship 

> We have to account for multiple callbacks registered with this.teardown and this.commit (naming TBD) right? this.teardown and this.commit can only accept one callback but they can be called multiple times (each call adding a callback). If a context extension uses this.teardown and a component uses this.teardown, we need to fire both callbacks synchronously on teardown. Just want to make sure I'm reading that right.

So it gets confusing because multiple calls can share the same callback so I think the best way to do it would be: when you call it to either method it adds the callback to a set of callbacks which all fire when the respective events occurs. When the respective events are fired, this callback set is cleared. So in essence, there is at most one call per call and callback per actual event. Let me know if this is unclear and I can explain a second time.

Also I’m not sure about teardown but I’m partial to `this.schedule`. The tricky thing with `this.commit` is that it implies that you’re actually committing the node rather than passing a callback which fires on commit. We need names that don’t imply action but reaction catch my drift?

ryhinchey:
> So it gets confusing because multiple calls can share the same callback so I think the best way to do it would be: when you call either method, they add the passed callback to an internal set of callbacks which all fire when the respective events occurs. When the respective events are fired, this callback set is cleared. So in essence, there is at most one call per call and callback per actual event. Let me know if this is unclear and I can explain a second time.

Makes complete sense

> Also I’m not sure about teardown but I’m partial to `this.schedule`. The tricky thing with `this.commit` is that it implies that you’re actually committing the node rather than passing a callback which fires on commit. We need names that don’t imply action but reaction catch my drift?

Yah sounds good 😀

brainkim:
@ryhinchey If this is a commitment to implement the above proposals let me know! I’m just as eager to have this functionality so people stop saying things like this when really I’m just sorta trying to iterate on design.

https://twitter.com/rauchg/status/1256798212666478592
Screen Shot 2020-05-04 at 12 02 08 AM

Closing PR let me know if you want it open for some reason.

ryhinchey:
@brainkim im going to start the implementation tonight.  I also saw that tweet.  If anything, Crank has turned some heads in the React community and that’s awesome.  Iterate away!

brainkim:
@ryhinchey Didn’t mean to pressure you! I just wanted to coordinate and I’m tired of remaining silent on hooks (I have nothing nice to say, but like to abide by the engineering principle that you shouldn’t be critical about a solution until you have concrete alternatives). Take the time you need and if you have any questions feel free to post here/discussions/email/discord (brain#4221).

ryhinchey:
No worries! FWIW  I have lots of nice things to say about hooks.  There are also a lot of options for what state/logic sharing can become with Crank.  Hooks made sense for React and we can take a wildly different approach.  Doesn’t mean one is right and one is wrong 🚀

PR #91, opened 2020-05-02
deps: setup husky

appear:
This PR is for adding husky and lint-stage.
add verification step before commit.
This library remove unnecessary mistakes.


brainkim:
I have vague memories of terrible experiences with husky/lint-staged (like most JS tooling) but don’t remember exactly what my problems with it were. I guess I should let myself be hurt again. Let’s do it, I guess, but I reserve the right to excise it if it gives me trouble during development. I commit to merge by the end of the weekend (some build issues we’re having are taking priority 1 right now).

appear:
Have a nice day ~~ 🚗

Issue #36, opened 2020-04-18
React hooks


lishine:
Currently a lot of business logic, functionality and libraries being written as custom hooks using hooks api.
Theoretically, would there be a way to use them with this framework?
Of course, maybe this generator way of writing components will enable to rewrite same logic in a much simpler way.
But I am asking about using existing code as it is.
For example, react-query, which exposes { isFetching, data } = useQuery(query, fetch query).

benjamingr:
Hooks are deeply ingrained into React and using them would basically be a reimplementation. 

This library can implement useState/useEffect in a way that lets you use hooks inside generators - though in order to keep that state internally you'd need an await:

```js
const { isFetching, data } = await useQuery(query, fetch query);
```

Or something similar - though to be fair I think the benefit of hooks here is marginal. In this library you can just use regular vanilla JavaScript functions.

The way you'd reuse logic from hooks is you'd take the non-frameworky parts of the code (the library part) and just use that. 

In your case - you'd just use graphql/apollo directly without having to go through a hook - which lets you use most of the code directly.

lishine:
I am thinking, if replacing each of these useEffect useState etc with crank equivalent, then such custom hooks could be used as is.
Then the adaption could be faster

workingjubilee:
What Brian might not say is: there is a danger in "can you make this work like in React", and the answer being "yes" too many times, and then people complain that the library does not work like React (because it is not), and then all the people who were interested in it being "React 2.0" leave, and all the people who didn't care about its similarity to React and wanted something new have already left.

I could see someone writing a hooks shim that lets some Hooks code get reused, but I think it would be dangerous for Brian to do so.

ryhinchey:
> What Brian might not say is: there is a danger in "can you make this work like in React", and the answer being "yes" too many times, and then people complain that the library does not work like React (because it is not), and then all the people who were interested in it being "React 2.0" leave, and all the people who didn't care about its similarity to React and wanted something new have already left.

+ 1 to this!

lishine:
> What Brian might not say is: there is a danger in "can you make this work like in React", and the answer being "yes" too many times, and then people complain that the library does not work like React (because it is not), and then all the people who were interested in it being "React 2.0" leave, and all the people who didn't care about its similarity to React and wanted something new have already left.
> 
> I could see someone writing a hooks shim that lets some Hooks code get reused, but I think it would be dangerous for Brian to do so.

[...]

PR #100, opened 2020-05-07
feat: adding oncommit and onteardown methods

ryhinchey:
This PR adds a few new methods to the Context class.  Two of them would be documented as part of Crank's public api:

`oncommit` - takes a callback that runs after your component returns or yields a value.   As noted here https://github.com/bikeshaving/crank/issues/33, this could be a good place to access ref values.  We can work through the ref api as part of this PR if we want.  I actually kind of like just passing `ComponentNode.value` to the oncommit callback; although, using ref callbacks and knowing to only access the values in your `oncommit` callback could work too.

`onteardown` - takes a callback that runs after your component is unmounted (removed from the dom). This is a great place to clear listeners.

Here are some examples of how you could use these methods:

```jsx
function *Counter() {
  let count = 0;

    count = count + 1;
    this.refresh();
  }, 500);


  this.onteardown(() => {
    clearInterval(interval);
  });

  // potential api passing ref values
  this.oncommit((refs) => {
    for (let el of refs) {
      console.log(el.offsetWidth);
    }
  });

  while (true) {
    yield <div><p>{count}</p></div>;
  }
}
```

TODO: add more tests for the `onteardown` method

### DIFF
--- a/src/__tests__/index.tsx
+++ b/src/__tests__/index.tsx
@@ -1764,3 +1764,97 @@ describe("parent-child refresh cascades", () => {
 		});
 	});
 });
+
+describe("oncommit", () => {
+	afterEach(async () => {
+		document.body.innerHTML = "";
+		await renderer.render(null, document.body);
+	});
+
+	test("sync function", () => {
+		const runOnCommit = jest.fn();
+
+		function Component(this: Context): Element {
+			this.oncommit(() => {
+				runOnCommit();
+			});
+
+			return <span>one</span>;
+		}
+
+		renderer.render(
+			<div>
+				<Component />
+			</div>,
+			document.body,
+		);
+
+		expect(runOnCommit).toHaveBeenCalledTimes(1);
+	});
+
+	test("async function", async () => {
+		const runOnCommit = jest.fn();
+
+		async function Component(this: Context): Promise<Element> {
+			this.oncommit(() => {
+				runOnCommit();
+			});
+
+			await new Promise((resolve) => setTimeout(resolve, 100));
+			return <span>one</span>;
+		}
+
+		await renderer.render(
+			<div>
+				<Component />
+			</div>,
+			document.body,
+		);
+
+		expect(runOnCommit).toHaveBeenCalledTimes(1);
+	});
+
+	test("sync generator", () => {
+		const runOnCommit = jest.fn();
+		let ctx!: Context;
+		function* Component(this: Context): Generator<Element> {
+			this.oncommit(() => {
+				runOnCommit();
+			});
+			ctx = this;
+			while (true) {
+				yield <span>Hello world</span>;
+			}
+		}
+
+		renderer.render(
+			<div>
+				<Component />
+			</div>,
+			document.body,
+		);
+
+		expect(runOnCommit).toHaveBeenCalledTimes(1);
+		ctx.refresh();
+		expect(runOnCommit).toHaveBeenCalledTimes(2);
+	});
+
+	test("async generator", async () => {
+		const runOnCommit = jest.fn();
+
+		async function* Component(this: Context): AsyncGenerator<Element> {
+			this.oncommit(async () => {
+				runOnCommit();
+			});
+
+			for await (let _ of this) {
+				yield <span>one</span>;
+			}
+		}
+
+		await renderer.render(<Component />, document.body);
+		expect(runOnCommit).toHaveBeenCalledTimes(1);
+		await renderer.render(<Component />, document.body);
+		expect(runOnCommit).toHaveBeenCalledTimes(2);
+	});
+});
--- a/src/index.ts
+++ b/src/index.ts
@@ -862,7 +862,7 @@ class ComponentNode<T, TProps> extends ParentNode<T> {
 			// TODO: batch this per macrotask
 			this.parent.commit();
 		}
-
+		this.ctx.callOnCommitCallbacks();
 		this.updating = false;
 		return; // void :(
 	}
@@ -874,6 +874,7 @@ class ComponentNode<T, TProps> extends ParentNode<T> {
 
 		this.updating = false;
 		this.unmounted = true;
+		this.ctx.callOnTeardownCallbacks();
 		this.ctx.clearEventListeners();
 		if (!this.finished) {
 			this.finished = true;
@@ -1014,10 +1015,15 @@ export class HostContext<T = any> {
 export interface ProvisionMap {}
 
 const componentNodes = new WeakMap<Context<any>, ComponentNode<any, any>>();
+const commitCallbacks = new WeakMap<Context<any>, Set<Function>>();
+const teardownCallbacks = new WeakMap<Context<any>, Set<Function>>();
+
 export class Context<TProps = any> extends CrankEventTarget {
 	constructor(host: ComponentNode<any, TProps>, parent?: Context<TProps>) {
 		super(parent);
 		componentNodes.set(this, host);
+		commitCallbacks.set(this, new Set());
+		teardownCallbacks.set(this, new Set());
 	}
 
 	/* eslint-disable no-dupe-class-members */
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -1045,6 +1051,32 @@ export class Context<TProps = any> extends CrankEventTarget {
...
+	}
+
+	callOnTeardownCallbacks(): void {
+		const callbacks = teardownCallbacks.get(this);
+		for (const callback of callbacks!) {
+			callback();
+		}
+
+		teardownCallbacks.delete(this);
+	}
```
brainkim, line 1079: Unless we can think of reasons for the user to call `callOnCommitCallbacks` or `callOnTeardownCallbacks`, it should not be exposed to the context. The Context is sorta like the place where all public methods go, so to put them here would be incorrect, and it could run into the user calling these methods and unexpectedly triggering callbacks when they weren’t supposed to happen.

src/index.ts
```diff
@@ -1014,10 +1015,15 @@ export class HostContext<T = any> {
 export interface ProvisionMap {}
 
 const componentNodes = new WeakMap<Context<any>, ComponentNode<any, any>>();
+const commitCallbacks = new WeakMap<Context<any>, Set<Function>>();
+const teardownCallbacks = new WeakMap<Context<any>, Set<Function>>();
```
brainkim, line 1019: These sets should probably be defined directly on the `ComponentNode` class, to reduce the number of weakmaps.

src/index.ts
```diff
@@ -862,7 +862,7 @@ class ComponentNode<T, TProps> extends ParentNode<T> {
 			// TODO: batch this per macrotask
 			this.parent.commit();
 		}
-
+		this.ctx.callOnCommitCallbacks();
```
brainkim, line 865: We might have to make sure to check `this.updating`. `this.updating` is a boolean which is set to true when a parent initiates an update, and set to false when the component finally commits. I need to think about whether we should fire callbacks here when a commit is initiated by a child, or when `this.refresh` is called on the current component.

src/index.ts
```diff
@@ -874,6 +874,7 @@ class ComponentNode<T, TProps> extends ParentNode<T> {
 
 		this.updating = false;
 		this.unmounted = true;
+		this.ctx.callOnTeardownCallbacks();
```
brainkim, line 877: I’m not sure about the ordering of this, whether it should happen here, after the generator associated with the component is returned, or after the children are unmounted. This is one of those design decisions that takes like a day of staring into space and walking in circles on a kitchen tile to figure out.

src/__tests__/index.tsx
```diff
@@ -1764,3 +1764,97 @@ describe("parent-child refresh cascades", () => {
 		});
 	});
 });
+
```
brainkim, line 1767: Maybe we can put these tests in their own file. I’m starting to break the files out individually by topic so we don’t just have a giant index.tsx test file.

src/index.ts
```diff
@@ -1045,6 +1051,32 @@ export class Context<TProps = any> extends CrankEventTarget {
 	refresh(): MaybePromise<undefined> {
 		return componentNodes.get(this)!.refresh();
 	}
+
+	oncommit(callback: Function): void {
```
brainkim, line 1055: Probably want to type these methods.

ryhinchey:
> I can be convinced otherwise, but part of me has a little bit of love for Yukihiro Matsumoto-esque APIs which look and sound like English.

I think `cleanup` will be pretty obvious to people.  Whether it's `commit` or `schedule`, people will have to learn what it means.  I'm down with naming it `schedule`.  It needs to be something that indicates your callback will fire once if using a sync/async function and 1+ times using a sync/async generator.  Schedule doesn't really get me there but I can't come up with something better at the moment.

When I work on the docs as part of this PR, I'll try my best to encapsulate what people can expect from `schedule` (or whatever the final name is).  A lot of people will be assuming it's like `componentDidMount` so I'll add how it's different and examples of how you can do `componentDidMount` in Crank.

src/index.ts
ryhinchey: I completely agree!  I was going to mention that Context should encapsulate the public facing api for components.  I saw `setDelegates` in Context though and figured that it was ok for now to put these there.  

@brainkim Do you agree that `setDelegates` should be removed from Context as well?

src/index.ts
ryhinchey: sounds good! 

@brainkim is there any reason the `componentNodes` weakmap isn't defined in the `Context` class?

brainkim:
> It needs to be something that indicates your callback will fire once if using a sync/async function and 1+ times using a sync/async generator. 

I think callbacks should fire at most once regardless of the kind of components they are because:

1. Callbacks which fire multiple times can be considered resources which need to be unregistered. These methods should be low level and not have to deal with deregistration.
2. Plugins should not need knowledge of what the specific component type is and differing behaviors would make this difficult.
3. A schedule which fired multiple times could be misused to create stateful components out of stateless functions. I really want people to rely on JS control flow as much as they can and only use these methods in helper functions or when absolutely necessary.

Lemme know if you want to pair or have questions. I should be free at some point today, tomorrow, this weekend.

src/index.ts
brainkim: Yes I do!

src/index.ts
brainkim: Not sure what you mean, code example?

src/index.ts
ryhinchey: I'd like to better understand the pros/cons of both.  Perhaps we can chat about this topic more

src/index.ts
ryhinchey: yah - let's set up some time to pair on it

ryhinchey:
> 3\. schedule which fired multiple times could be misused to create stateful components out of stateless functions. I really want people to rely on JS control flow as much as they can and only use these methods in helper functions or when absolutely necessary.

my thinking was that schedule fires any time a component returns or yields.  if generators yield more than once, so would `schedule`

src/index.ts
ryhinchey: https://github.com/bikeshaving/crank/blob/master/src/index.ts#L1016

src/index.ts
ryhinchey: > Probably want to type the callbacks to these methods.

for sure

> Another thing I thought about, what if onteardown or whatever took a possible async function?

i dont see why someone couldn't!

ryhinchey:
I ended up being pretty busy last week and couldn’t get to this.  Will be working on it this week

brainkim:
@ryhinchey Might hold off on it I’ve been a very naughty boy and experimented with some big refactorings for performance, mostly unsuccessfully that are going to be hard to rebase against.

brainkim:
@ryhinchey I hate to do this, but I think I might want to undelegate this feature. #33, #36 and #82 have been burning a hole in my brain for the past two weeks, and I think I see the path forward in terms of design and implementation. I want to get at least one new feature in the 0.2 version besides changing builds and supporting more environments, and I want 0.2 to be released soon (like maybe today or tomorrow) so I can submit Crank to the js-frameworks-benchmark competition and see how it actually does on someone else’s computer.

I know, I know, I should delegate more work so that Crank truly becomes an open source project, and I really do want to decrease the bus factor of the codebase. I deeply apologize if I’m stepping on your work, and I know it must feel unfair because I’m sorta just shipping releases without any sort of code review while other people have to deal with my captious reflections. It’s also unfair because I have the entire Crank codebase and all of its unexplained boolean flags in my head, and I need to do a better job of making the codebase understandable. At the same time, index.ts is currently shifting rapidly; for instance, I think that for performance reasons we might be moving away from inheritance for the different node types in favor of a single node type, with internal functions which operate on them.

Again, I’m really sorry Ryan! As an aside, I guarantee and promise you that I won’t be looking at or touching the internal event system until at least June 15th,  if that’s something that you want to look into.

ryhinchey:
No worries at all @brainkim!  Feel free to close this PR or amend it.  I’ve been busy on other things the past week and a half anyways.  No bad vibes 😁

ryhinchey:
Closing in favor of https://github.com/bikeshaving/crank/pull/115

brainkim:
Thanks Ryan!

PR #101, opened 2020-05-07
Added some missing URLs to package.json

mcjazzyfunky:
Some important URLs are missing in package.json, which means that those URLs are not shown on https://www.npmjs.com/package/@bikeshaving/crank.
I've added those URLs to "package.json".


brainkim:
Thank you!

PR #107, opened 2020-05-15
Esnext esm rollup, optionally w/o EventTargetShim

### LINKED ISSUE #65: Esnext esm/* target, rather than es5?
Package of `@bikeshaving/crank` 0.1.1 at npm has a 142k esm/ dir. Odd because github files are much smaller, 42k only (10k gzip) for all direct src/ children. From what I see in esm/index-4b79835a.js, the build system adds huge polyfills for very old browser engines, and replaces all usage of generators and async with es5 state machines. Since these files are es modules which `export` stuff, one can't just `require` them in old node or run in browser as `script` without an additional transpilation step on behalf of the app developer. My question is, how does applying such radical transforms to the esm dir help users of old engines in practice then? Doesn't it negate the top benefit of crank, excellent runtime performance in modern browsers due to using their native features? Would reconfiguring the build system to keep esnext generators and async in the esm dir make sense?

nykula:
Split cjs,esm plugin rollup config. For esm, override tsconfig target compiler option with esnext. For both, replace EventTargetShim d.ts import with DOM EventTarget in the generateBundle hook. Move event-target-shim back to build dependencies.

If EVTSHIM=0 rollup, use native EventTarget only without importing event-target-shim or inlining it in the js. Dummy EventTarget=Object if undefined, so that serverside import doesn't crash (I split fetch and render, not using events there). This results in 30% smaller file, winning TTI milliseconds on embedded devices where I build engines from source and so know they're modern enough.

Size of esm is 184k before (commit 003d2c2), 152k after, 92k with EVTSHIM=0. Resolves #65. Please build and try this in some real-world app copy before merging, because my app is still Preact-based, I only ran this against the few hundred lines of small components I could port to Crank quickly.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,18 +1,36 @@
 import resolve from "@rollup/plugin-node-resolve";
 import typescript from "rollup-plugin-typescript2";
-export default {
-	input: ["src/index.ts", "src/dom.ts", "src/html.ts"],
-	output: [
-		{
-			format: "cjs",
-			dir: "cjs",
-			sourcemap: true,
-		},
-		{
-			format: "esm",
-			dir: "esm",
-			sourcemap: true,
-		},
-	],
-	plugins: [typescript(), resolve()],
-};
+function generateBundle(_, bundle) {
+	for (var f in bundle)
+		if (bundle[f].fileName.endsWith(".d.ts"))
+			bundle[f].source = bundle[f].source
+				.replace(/.*event-target-shim.*\s*/, "")
+				.replace(/Shim implements EventTarget/, "");
+}
+export default [
+	{
+		input: ["src/index.ts", "src/dom.ts", "src/html.ts"],
+		output: [{format: "cjs", dir: "cjs", sourcemap: true}],
+		plugins: [typescript(), resolve(), {generateBundle}],
+	},
+	{
+		input: ["src/index.ts", "src/dom.ts", "src/html.ts"],
+		output: [{format: "esm", dir: "esm", sourcemap: true}],
+		plugins: [
+			typescript({tsconfigOverride: {compilerOptions: {target: "esnext"}}}),
+			process.env.EVTSHIM !== "0" && resolve(),
+			process.env.EVTSHIM === "0" && {
+				transform(code) {
+					return {
+						code: code.replace(
+							/.*event-target-shim.*/,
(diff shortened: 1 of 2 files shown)


brainkim:
Hmmm the big problem with optionally requiring event-target-shim is that safari just doesn’t have a working EventTarget class right now 🙃 I think the best solution is to leave it as is and write a clean room implementation of event targets which also implement features like custom capturing/bubbling and event delegation. Lots of work to do!

The other optimizations will definitely be required and I’m planning to get it all in in a 2.0 release. Thanks for trying Crank! Let me know your thoughts in a discussion or here or whatever.

ryhinchey:
We should be able to keep the other parts of this PR that don’t touch event target shim.  @brainkim i agree we should keep that as is for now but transpiling the esm build with an esnext target makes sense to me

rollup.config.js
```diff
@@ -1,18 +1,36 @@
...
+	{
+		input: ["src/index.ts", "src/dom.ts", "src/html.ts"],
+		output: [{format: "cjs", dir: "cjs", sourcemap: true}],
+		plugins: [typescript(), resolve(), {generateBundle}],
+	},
+	{
+		input: ["src/index.ts", "src/dom.ts", "src/html.ts"],
+		output: [{format: "esm", dir: "esm", sourcemap: true}],
+		plugins: [
+			typescript({tsconfigOverride: {compilerOptions: {target: "esnext"}}}),
```
brainkim, line 20: Ohhh so this is how you do it. Is it just me or is rollup-plugin-typescript2’s documentation completely lacking for figuring out this basic detail

ryhinchey: I honestly tried doing this for around 2 hours and couldn’t get the overrides to work and left frustrated.  Glad to see @nykula figured it out!  I too found the docs lacking around this

brainkim:
Closing in favor of #108. I do want to remove event-target-shim from our dependencies, but it is not gonna happen yet.

My current thinking is that we’re going to ship ESNext everywhere because it’s better for performance and if your environment doesn’t support async/generator functions anyways you’re gonna have a hard time writing Crank components to begin with, and as a framework we can require transpilations of dependencies for older environments.

Thanks for the effort though! I do plan on making some changes to reduce crank bundle size soon!

nykula:
@brainkim @ryhinchey Ok! I appreciate the review.

PR #118, opened 2020-05-21
Context refactor

ryhinchey:
This PR does a small refactor to `Context`.  It was mentioned before that Context should be considered a public api.  There were two functions in particular that didn't need to be there related to delgates which are removed.  In addition, the event emitter code has been moved directly under the `Context` class.

- the `Context` class is now in its own file
- `setDelegate` and `setDelegates` are under `ComponentNode`
- `events.ts` file has been removed and event emitter stuff is under `Context`


brainkim:
Closing in favor of https://github.com/bikeshaving/crank/pull/123. Sorry again, Ryan! Starting with 0.2.0 the event stuff is all handled via private internal functions.

brainkim:
Also, one note on the PR. I’m a big fan of having things in a single file. The Crank.js core is a single file! It’s a weird flex but I’m okay with it! Libraries like Backbone.js and Underscore.js all used to be single files, and I’m happy to continue in that tradition.

PR #123, opened 2020-05-26
Reducing runtime memory cost, execution time and bundle size

brainkim:
**Note: it will be more or less impossible to rebase off this branch if you had changes in 0.1.x. Sorry! Hopefully churn will decrease soon.**
The last set of changes I want to make before the 0.2.x release.
- [x] yeet linked list implementation into space
  - [x] yeet LeafNode into space
  - [x] use the zero-one-many optimization to avoid array allocations
- [x] stop using inheritance, have a single node class rather than 3/4
  - [x] yeet ComponentNode into space
  - [x] yeet HostNode into space
  - [x] yeet FragmentNode into space
- [x] use a bitmask for node flags
- [x] stop using the iterable context pattern in intrinsics
  - [x] pass HostNode to intrinsics directly
  - [x] yeet HostContext into space
- [x] remove redundant/computable properties from ~~Nodes~~ Elements
  - [x] intrinsic
  - [x] childScope
  - [x] childValues
  - [x] dirty flags
    - [x] expose computed getters rather than flags directly
  - [x] clock
  - [x] renderer
  - [x] ctx (removed from intrinsic elements)
  - [x] scope
  - [x] parent
  - [x] delegates
  - [x] flags (move to contexts???)
  - [x] childrenByKey
- [x] use functions rather than methods (should help with mangling method names)
  - [x] functionify Node
  - [x] functionify renderer
  - [x] maybe it actually makes sense to use methods and not functions 😓
- [x] make elements and nodes the same thing (the thing that Inferno.js does)
  - [x] separate function for mounting
  - [x] pass renderer in rather than storing it on nodes
- [x] clean up the logic and naming in `updateChildren`
- [x] refactor update in the case of element reuse
- [x] kill event-target-shim (you were a great shortcut but it’s time for you to go)
- [x] refactor to a generator-free Renderer API for host elements
- [x] make commit, update and updateChildren return useful values
  - [x] fix types
  - [x] figure out why tests are failing in node 10
- [x] create a separate codepath for mounting children
- [x] get rid of remove method for now
- [x] Fix copies
  - [x] fix copies of async elements
  - [x] call ref functions on copies
- [ ] TESTS TESTS TESTS
  - [ ] error
  - [ ] portal
  - [x] async copy
  - [ ] portal copy
  - [ ] ref of portal
  - [x] ref of copy
  - [ ] EventTargets
    - [ ] once
    - [ ] portals
    - [ ] copies
    - [ ] raw
    - [ ] fragments
  - [ ] get some code coverage stuff going
- [x] mangle private properties


brainkim:
Too lazy to test more things right now.

PR #134, opened 2020-07-20
Improve error handling and recovery

brainkim:
Fixes some edge cases and makes errored components slightly more recoverable


brainkim:
Currently we mark generator components which directly throw an error as finished; in other words we treat them the same as generator components which return. Maybe we could consider resetting the components at some point so that they can be rerendered without being unmounted. Maybe even have another method that allows resets too. Something to think about, but probably not for the next release.

Issue #145, opened 2020-08-05
root instanceof Node returns false after upgrading from 0.1.6


calumsult:
To preface, I think this project is really neat, and I'm probably doing something totally wrong. 👍 

I made a fun test project with 0.1.6 and it worked great. I recently updated it to use current (0.3.1) and doing something as simple as

`renderer.render(<div>hello</div>, document.body);`

errors with root is not a node. I'm using parcel. If I add the crank source directly to the project it works (root instanceof Node now returns true). Either way, if I check root instanceof Node before calling render it always returns true, it's only inside the cjs version in the npm package that it returns false. Any thoughts on what I'm doing wrong?

Thanks for your time!

brainkim:
Hmm, if document.body is not a node, it’s usually null. This happens because the script tag is written in the head. You can fix this by making the script tag async or listening to an event like DOMContentLoaded. This is a good reference for how script tags work (https://eager.io/blog/everything-I-know-about-the-script-tag/).

If this isn’t the problem, lemme know! I’m happy to debug but I’d need a little more in the way of code examples. Thanks for trying Crank!

brainkim:
Fixed by #147 and deployed in 0.3.3

PR #147, opened 2020-08-09
Check for properties rather than using instanceof

calumsult:
I worked out the issue with [#145](https://github.com/bikeshaving/crank/issues/145). I'm using nw.js, which runs node modules in a different context from the "browser". Thus instanceof fails. This is the same if you were using elements in an iframe.

https://docs.nwjs.io/en/latest/For%20Users/Advanced/JavaScript%20Contexts%20in%20NW.js/#problem-with-instanceof

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof

Would you consider swapping instanceof for duck typing to test an instance type? I'm not sure of the best way to do that (I can research and submit a second pull request if this is something you will support); this pull request is just a rough guess.

nw.js is a bit annoying for these edge cases, but it's really fun to quickly roll applications, and crank pairs nicely with this use case. I really like what you've built.

Thanks!

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -15,7 +15,7 @@ export class DOMRenderer extends Renderer<Node, string | undefined> {
 		root: Node,
 		ctx?: Context,
 	): Promise<ElementValue<Node>> | ElementValue<Node> {
-		if (!root.nodeName) {
+		if (!(root != null && typeof root.nodeType === "number")) {
 			throw new TypeError(
 				`root (${root && (root as any).toString()}) is not a node`,
 			);
@@ -69,7 +69,7 @@ export class DOMRenderer extends Renderer<Node, string | undefined> {
 	}
 
 	patch(el: CrankElement<string | symbol>, node: Element): void {
-		const isSvg = node.toString() === "[object SVGElement]";
+		const isSvg = node.namespaceURI === SVG_NAMESPACE;
 		for (let name in el.props) {
 			let forceAttribute = false;
 			const value = el.props[name];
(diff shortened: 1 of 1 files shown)


brainkim:
Ah I see. Let me do some research to see if there are more bullet-proof/performant ways to identify nodes and SVG elements, but I am on board with duck-typing.

calumsult:
Excellent, thanks!

brainkim:
Okay. Did a little searching and noodling, and here’s what I think.

For the node check I think it should probably be something like `node != null && typeof node.nodeType === "number"`. We need to make sure we check null/undefined so the error is as expected, and typeof checks are a little faster and more accurate. We could also add a check for `nodeName` but this node check is mainly to catch errors early on so I’m not too worried about perfect accuracy.

The SVG check is a little more important and luckily the DOM [has a helper property](https://developer.mozilla.org/en-US/docs/Web/API/Element/namespaceURI) to determine the namespace. It should be `node.namespaceURI === SVG_NAMESPACE`.

brainkim:
@calumsult Any updates? I can do this myself I just wanted to make sure the code worked for your nw.js use-case.

calumsult:
Sorry, I got sidetracked, I will test it this weekend.

calumsult:
`node != null && typeof node.nodeType === "number"` and `node.namespaceURI === SVG_NAMESPACE` work. 👍

brainkim:
Thanks! I’ll try to get this deployed ASAP.

brainkim:
Deployed in 0.3.3

PR #156, opened 2020-09-18
Add export to ChildIterable

toddlucas:
This fixes TypeScript error TS4082 that occurs when exporting a `Children` type that gets expanded into `Child | ChildIterable` by the compiler.

I'm not sure exactly what causes TS to expand the type in this case. It occurs when exporting a [connected component](https://github.com/toddlucas/crank-redux/blob/190629c3eac93e3b886fed114081b992750dcdce/src/index.ts#L30) in a new experimental library, based on the generated `.d.ts` declaration.

```
const MyComponent = () => createElement('div');
export default connect(MyComponent);
```

### DIFF
--- a/src/crank.ts
+++ b/src/crank.ts
@@ -101,7 +101,7 @@ export type Raw = typeof Raw;
 export type Child = Element | string | number | boolean | null | undefined;
 
 // NOTE: we use a recursive interface rather than making the Children type directly recursive because recursive type aliases were added in TypeScript 3.7.
-interface ChildIterable extends Iterable<Child | ChildIterable> {}
+export interface ChildIterable extends Iterable<Child | ChildIterable> {}
 
 /**
  * Describes all valid values of an element tree, including arbitrarily nested iterables of such values.


brainkim:
Looks good. Just to be clear, what is the full error you’re seeing/version of TypeScript. I know that these type is not exported errors happen just want to make this PR more easily searchable.

brainkim:
Deployed as 0.3.5. Thanks and happy experimenting!

toddlucas:
The return type of the `connect` method is defined as

```
function connect(...): (this: Context<TOwnProps>, props: TOwnProps) => Iterator<Children, Children | void, any>;
```

with the `Iterator` result being equivalent to the generator-specific result type. However, it's resolved to an expanded form (with the second `Children` generic argument term being expanded):

```
Iterator<Children, string | number | boolean | void | Element<Tag> | ChildIterable | null | undefined, any>
```

which may be due to unresolved generic type parameters on `connect`. The error message is:

```
error TS4082: Default export of the module has or is using private name 'ChildIterable'.
```

Thanks!

brainkim:
Yes that’s a weird TypeScript error which I’ve never understood 😓 Thanks again!

PR #174, opened 2020-11-15
Typo in lifecycle guide

neonfuz:
Just a small typo I noticed in the guide. Probably at some point the code example was changed without changing the text referencing it.


brainkim:
Thanks! I’ll try to deploy soon.

Issue #204, opened 2021-05-11
Removing a `type` prop from an `<input>` doesn't remove the type from the DOM node


GormanFletcher:
[Reproduction](https://codesandbox.io/s/crankjs-unrendered-view-repro-forked-qgyb0?file=/src/index.tsx)

If I have an input with `type="email"` and I replace it with an untyped input, the input in the DOM retains the `type` attribute.

GormanFletcher:
Ah! I forgot that from [the docs](https://crank.js.org/guides/handling-events#form-elements).

For implementation, would it make sense to diff the attrs on the reused DOM node against the new render's props, from the DOM instead of tracked state?

I don't understand if that would prompt API changes - if that puts us all the way in controlled/uncontrolled territory. The current behavior is a tricky caveat to "declarative", coming from other frontend frameworks, and its effects can be subtle to detect, since carryover attrs won't always produce easily visible changes (like an input that looks correct but fails to validate).

An explicitly `null` prop works in small examples like your comment or the docs, where logically we're rendering different versions of the same semantic element. I stumbled on this with a more complex example: a login/signup component that shows/hides the extra signup fields, which happen to be structurally similar enough to cause DOM node reuse. The signup form prompts name, then email, while the login form prompts just for email. In my mind, I'm declaring that the name input "goes away" and the email field that's identical in both code paths stays as-is. I'm not mentally modeling it as a mutation of one single element.

Maybe the answer is to explicitly tell Crank about that mental model? Liberal application of `crank-key`? That fixes the behavior in my codesandbox, and means I don't have to think about how code paths interact under the hood. But choosing that solution, I'd wind up peppering it all over the place defensively without knowing if I really need it everywhere, which doesn't feel great. So I don't know.

Thoughts?

GormanFletcher:
I overlooked a complicating factor: because my `<input>` is uncontrolled, when my component toggles and Crank mutates the `<input>` in-place, the input will retain anything the user has typed, even if that value is inappropriate for the new meaning of the `<input>`. Deleting attrs won't help that - it sounds like I definitely need to use `crank-key` in that case.

Needing to watch out for attr preservation still strikes me as tricky, but now I don't have a concrete use case to point to for it, so take that as you will.

brainkim:
[...]

Issue #180, opened 2020-12-22
crank-ref cannot focus element synchronously


spiffytech:
# 🐛 bug report

I have a component that includes a `<textarea />` that I want to focus on creation. If I use a `crank-ref` callback, the element does not focus unless I wrap the callback in a `setTimeout`.

# 💻 Code Sample

## Doesn't work

```typescript
function TextAreaTest() {
  return (
    <div>
      <!-- lots o' code here -->
      <textarea crank-ref={(el: HTMLTextAreaElement) => el.focus()} />
    </div>
  );
}

renderer.render(<TextAreaTest />, document.body);
```
## Works

```typescript
function TextAreaTest() {
  return (
    <div>
      <!-- lots o' code here -->
      <textarea crank-ref={(el: HTMLTextAreaElement) => setTimeout(() => el.focus(), 0)} />
    </div>
  );
}

renderer.render(<TextAreaTest />, document.body);
```

brainkim:
Hmmm this is probably because of ref callbacks firing a little too early. They seem to be firing after DOM node creation rather than when they’re inserted into the parent. I will think about what we can do, and maybe overhaul the timings of these callbacks, but in the meantime you can continue using a `setTimeout()` or `requestAnimationFrame()` callback to mitigate this problem.

brainkim:
Also, for this specific use-case, you may wish to try the ]`autofocus` attiribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus) instead.

waynebaylor:
I noticed the same behavior when yielding and using `schedule(..)`.

This didn't work because the node was not yet connected to the dom:
```javascript
let node = yield <input ... />;
node.focus(); // node isn't connected to the dom yet
```

Also, this didn't work for the same reason:
```javascript
this.schedule(node => {
  node.focus(); // node isn't connected to the dom yet
});
yield <input ... />;
```

To work around the timing issue I chose to monitor the `isConnected` property:
```javascript
this.schedule(node => {
  const handle = setInterval(() => {
    if(node.isConnected) {
      node.focus();
      clearInterval(handle);
    }
  }, 100);
});
yield <input ... />;
```

brainkim:
@waynebaylor That’s a resourceful but non-ideal workaround! We don’t want to have to poll the node until it is actually inserted 😓

TL;DR: I’ll try to add a new callback method called `complete()` which fires when nodes are rendered into the DOM.

[...]

Issue #175, opened 2020-11-16
Boolean attributes/properties like spellcheck.


brainkim:
A funny bug. The `spellcheck` property is a boolean, but the `spellcheck` attribute is either the `"true"` or `"false"` string. When attempting to set the `spellcheck` property to the `"false"` string, it gets set to true because this string itself is truthy. I think we might need a better global list of which props should be dealt with as properties versus which props should be dealt with as attributes. If you know a library which has a good global list of these props/attributes lemme know.

In the meantime, if you encounter a bug like this you can use `spellcheck={false}` to set the boolean property directly.

brainkim:
This file from React seems to have a wealth of information about various attributes. https://github.com/facebook/react/blob/master/packages/react-dom/src/shared/DOMProperty.js

PR #208, opened 2021-07-10
Refactorings and features

brainkim:
Refactorings to the custom renderer API in preparation of deferred mount/unmounting of elements. Lots of inlining of functions whose structure didn’t really make much sense. Like why did we call a generic `commit()` from `updateChildren()`, when within the `commit()` function we would again start doing different logic based on committing element tag.

Beeg changes:
- [x] Overhaul of the internal renderer API (yet again).
  - [x] Stop passing elements into Renderer methods.
  - [x] Disappear the janky `Element.hadChildren` computed property.
  - [x] Get rid of element flags.
  - [x] Pass old props/children to the patch and arrange methods, in preparation for solving #204 and going back to a nicer story for removed properties.
  - [x] Rename `complete()` to `flush()`, and fix some bugs created in (#206) method for (#180).
  - [x] Move the internal renderer methods onto a `RendererImpl` interface, which is passed to the constructor.
  - [x] Consider reordering `node` and `tag` parameters in `patch()`,`arrange()` etc.
  - [x] Make `patch()` work per individual prop?
- [x] Clean up the internal functions to support future changes.
  - [x] Kill `updateChildren()`.
    - [x] Replace `updateChildren()` with a `diffChildren()` function which doesn’t do the committing part.
  - [x] Kill `commit()`.
    - [x] Move `commitCtx()` out of commit.
  - [x] Kill `update()`
    - [x] Don’t call `update()` from `Renderer.render()`.
    - [x] Move Fragment updating logic out of `update()`.
  - [x] Cache parses for Raw elements.
- [x] Create a “retainer” class to store all the internal state which we’re currently storing on elements.
- [x] Consider using the newly retained childValues cached property in more places.
  - [x] Store the old childValues on component contexts.
  - [x] Avoid rearranging when components produce the same nodes.
  - [x] Use cached childValues in getChildValues/getValue functions.
- [x] DOM renderer upgrades
  - [x] #204
    - [x] Use the `Copy` tag for new uncontrolled behavior???????
  - [x] #175
  - [x] ~~#209~~
- [x] ~~`crank-static`~~ (TODO in a separate PR)
- [x] ~~Async mounting and unmounting~~ (TODO in a separate PR)


brainkim:
Branch name is inaccurate but GitHub is scuffed and won’t let me change it.

brainkim:
https://twitter.com/jaredpalmer/status/1143112897553424384?lang=en

PR #230, opened 2022-05-12
v0.5.0

brainkim:
### Breaking changes

- The internal `RendererImpl` method `escape()` has been renamed to `text()`.
- The internal `RendererImpl` method `parse()` has been renamed to `raw()`, and the value `prop` is always passed into this method, regardless of whether the value is a string.
- Generator components which return will now restart and re-run rather than staying stuck on the returned value. This should help debugging components which have accidentally returned.

### Quality of life improvements
- Special props like `crank-key`, have an additional `$`-prefixed variant. Going forward, this will be the preferred syntax, but `crank-key` and `c-key` will continue to be supported as well.
	```ts
	<div $key={key} $ref={ref} $static />
	```
- Generator components which are in a `for...of` or `for await...of` loop will now
attempt to exit normally, so that cleanup code can be placed after the loop.
	```ts
	// before
	function Counter() {
		let i = 0;
			i++;
			this.refresh();
		}, 1000);
		try {
			for ({} of this) {
				yield <div>{i}</div>;
			}
		} finally {
			clearInterval(interval);
		}
	}

	// after
	function Counter() {
		let i = 0;
			i++;
			this.refresh();
		}, 1000);
		for ({} of this) {
			yield <div>{i}</div>;
		}
		clearInterval(interval);
	}
	```
- Async generator components can now use for...of loops. Async generator components which yield from a for...of loop behave like sync generator components, pausing at each yield.
	```ts
	async function Counter() {
		let i = 0;
			i++;
			this.refresh();
		}, 1000);
		for ({} of this) {
			yield <div>{i}</div>;
		}

		clearInterval(interval);
	}
	```
- The `Context` type can now be passed a function type as its first parameter to strongly type the `this` props iterators.
### New Features
- The automatic JSX runtime transform is now supported.
- Hydration: The DOM Renderer now provides a `hydrate()` method which will attempt to re-use DOM nodes already found on the page. A corresponding `hydrate()` method has been defined for custom renderers.
- Crank now provides a tagged template function called `jsx` which replicates JSX syntax and allows templates to be written with vanilla JavaScript.
	```ts

	```
- Calling `dispatchEvent()` on a component context will now trigger any `onevent` style props.

### Bug Fixes
- Errors which are thrown in Context event listener functions will not prevent other listeners from being called.
- Async generator components should run more predictably.
- IFrame/IMG `src` and other props which are sensitive to being re-assigned will not be re-assigned.
- `$static` elements which are re-rendered with different tags will now correctly re-render.


darrylcousins:
Hi Brian,

Good for you, sincerely; good for you. I've been coding javascript off and on since 1999 and the early days of DHTML. As far as I know this is the by far the easiest package to use for rendering and refreshing the DOM (React is a chore). Congratulations.

I've been using @bikeshaving/crank for a live site for 2 years, and at moment further extending the application.

I'm fear I'm still at `npm @bikeshaving/crank --version`  8.5.5 - I really should update.

You got this.

Ngā mihi nui,
Darryl Cousins

brainkim:
@darrylcousins Thank you so much for these kinds words, Darryl! I’m going through a difficult time and this is just the motivation I need to believe in myself. If we should ever meet in person, perhaps if in my journeys I end up in Aotearoa somehow, please let me buy you a drink or whatever.

src/template.ts
```diff
@@ -0,0 +1,347 @@
+import {c, Element} from "./crank.js";
+import type {Tag} from "./crank.js";
+
+export function t(
```
ZebulanStanphill, line 4: Funnily enough, in a private project I've been working on lately, I created a tagged template wrapper for the `__` localization/translation function in WordPress, to make value substitution easier compared to having to `sprintf` every time. I decided to call it `t` for "translate".

I'm guessing if WP ever did something similar, they'd probably just reuse `__` or go with something else, so I don't think it's that important. Just thought I'd point out that there is at least one other potentially obvious meaning that could be assigned to `t`. But I'm sure the same could be said for most single-letter names.

brainkim: I’m open to naming suggestions! I think taking a one-letter identifier is justified for a “framework,” and insofar as it could potentially be the single most called function in the library, I think the shorter the better.

How I got to `t`: I thought `jsx`/`tsx` might be misleading, given the divergences with `jsx` syntax extension. `htm`/`html` would have been even more misleading, given that JSX and HTML have only a familial resemblance. I thought maybe `x` would a nice alternative as in “XML,” and now I’m leaning towards `t` for `template`. Truly the most boring name to use, and I think `t` is easier to type out than `x`. Not really sure. Just giving you some insight into the design process.

This is kind of important because devtools will often read template tags without reading the import, so for example, a lot of editors come with automatic highlighting for `css`, `gql`, `html` template tags.

We could get cute and do something like `cog` or something. I dunno.

ZebulanStanphill: `t` being easy to type is definitely a great argument for using it over something like `x`. My only real issue with the former is that the term "template" is technically a bit vague. Come to think of it, tagged templates are literally already called "templates" regardless of their function, so I guess there's potential for a bit of confusion in terminology there.

How about `m` for "markup" instead? It's just as easy to type... in fact, I'd argue it's easier. It's also more specific in meaning.

PR #231, opened 2022-06-22
Fix how read-only DOM attributes are set

jake-low:
Hello!

I've been using Crank for a small project (delightful, by the way!) and found a bug with how SVG attributes are being set during DOM patching. I discovered this bug when I wrote code like this:

```jsx
renderer.render(
  <svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
    <rect x="0" y="0" width="10" height="20" class="foo" data-bar="1" />
  </svg>
);
```

and found that this didn't produce the desired DOM state, but instead something like this:

```html
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
  <rect class="foo" data-bar="1"></rect>
</svg>
```

where the x, y, width, and height attributes are missing.

I read the patch implementation and figured out the cause. The intent of the code as currently written appears to be this: try writing to the node property (`node[name] = value`), and if that throws an error, fall back to setting as an attribute (`node.setAttribute(name, value)`).

However, it seems based on my ad-hoc testing, attempting to set a read-only property on a DOM element (either an HTMLElement like `<p>` or an SVGElement like `<rect>`) doesn't actually throw an error; rather the assigned value is silently discarded.

```js
let p = document.createElement("p");
p.id = "foo"; // id is a writable property
console.log(p.id); // => foo

p.isContentEditable = true; // isContentEditable is a read-only property
console.log(p.isContentEditable); // => false

// same story but for SVG
let rect = document.createElementNS("http://www.w3.org/2000/svg", "rect")
console.log(rect.x);
// => SVGAnimatedLength {
// animVal: SVGLength {unitType: 1, value: 0, valueInSpecifiedUnits: 0, valueAsString: '0'}
// baseVal: SVGLength {unitType: 1, value: 0, valueInSpecifiedUnits: 0, valueAsString: '0'}
// }
rect.x = 10; // x is read-only
console.log(rect.x); // same as above
```

The solution I came up with is this: instead of `try`ing to set the property and falling back to `setAttribute` on error, instead walk up the object's prototype chain to find the owner of the property, and then get the descriptor of the property and use that to check if it's writable or not.

I added some additional test assertions for this, but these tests wouldn't actually have caught the bug, because JSDOM appears to permit writing to any Element property.

Feel free to let me know if you'd prefer a different approach; I'm happy to rework the code if you have feedback.

Thanks for your time, and for sharing Crank! It's really been a pleasure to use, and I really appreciate the care that was put into the docs and the motivation blog post.

### DIFF
--- a/src/__tests__/svg.tsx
+++ b/src/__tests__/svg.tsx
@@ -139,7 +139,7 @@ describe("render", () => {
 	test("non-string values", () => {
 		renderer.render(
 			<svg xmlns="http://www.w3.org/2000/svg">
-				<rect class="rectClass" x={10} y={20.5} width={5_000} height={null} />
+				<rect class="rectClass" x={10} y={20.5} width={5000} height={null} />
 			</svg>,
 			document.body,
 		);
--- a/src/__tests__/svg.tsx
+++ b/src/__tests__/svg.tsx
@@ -35,10 +35,33 @@ describe("render", () => {
 			</svg>,
 			document.body,
 		);
+
+		let svgRoot = document.body.firstChild;
 		expect(document.body.firstChild).toBeInstanceOf(SVGElement);
-		for (const child of Array.from(document.body.firstChild!.childNodes)) {
-			expect(child).toBeInstanceOf(SVGElement);
-		}
+
+		let rect = svgRoot!.childNodes[0] as SVGElement;
+		expect(rect).toBeInstanceOf(SVGElement);
+		expect(rect.tagName).toEqual("rect");
+		expect(rect.getAttribute("width")).toEqual("100%");
+		expect(rect.getAttribute("height")).toEqual("100%");
+		expect(rect.getAttribute("fill")).toEqual("red");
+
+		let circle = svgRoot!.childNodes[1] as SVGElement;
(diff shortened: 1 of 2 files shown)


jake-low:
Looks like I used some syntax that wasn't supported in Node 10; should be fixed now.

brainkim (approved):
@jake-low Thank you for this, and sorry about the late response. As always, I am both embarrassed and apologetic about such a low-hanging bug, likely made when I was trying to optimize the shit out of Crank. I could have sworn writing to a read-only property threw a hard error but here we are.

I continue to be annoyed by the lack of real DOM testing in the Crank.js test suite, and appreciate that you added tests which would fail, and plan on investigating migrating to a unit test library which works against a real DOM. If you have any recommendations let me know!

The code itself looks great! While I am slightly worried about performance, I think it’d probably be better to make things work first, and then worry about that stuff later. Thanks for debugging this, and I’ll have a release ASAP once I remember how to do this 😊😬❤️

brainkim:
Shipped in 0.4.3. Thanks again!

Issue #236, opened 2022-08-04
Uncaught TypeError: Invalid value used as weak map key


darrylcousins:
Hi Brian,

I was surprised that a crank app suddenly failed with a new Shopify theme. Removing the theme javascript solved the problem but of course is not the solution. The app is compiled with [Vite](https://vitejs.dev) and minified. Resulting in a show stopping error and the app not rendering:

```
boxesapp.js?v=160012563599259014851659593001:1 Uncaught (in promise) TypeError: c.then is not a function
    at Ot (boxesapp.js?v=160012563599259014851659593001:1:5435)
    at _e (boxesapp.js?v=160012563599259014851659593001:1:3873)
    at Ot (boxesapp.js?v=160012563599259014851659593001:1:5373)
    at _e (boxesapp.js?v=160012563599259014851659593001:1:3873)
    at oe (boxesapp.js?v=160012563599259014851659593001:1:9132)
    at boxesapp.js?v=160012563599259014851659593001:1:11469
    at async Promise.all (index 0)
    at async Object.fn [as init] (boxesapp.js?v=160012563599259014851659593001:7:10767)
```

Recompiling and disabling minify; the app will load correctly but with errors in the inspector:

```
Uncaught TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at addEventListener (boxesapp.js?v=181202394000030643281659593075:993:20)
    at Function.bindReady (jquery.min.js:2:9844)
    at init.ready (jquery.min.js:2:8340)
    at new init (jquery.min.js:2:7732)
    at e (jquery.min.js:2:6172)
    at jquery.min.js:2:19170
    at jquery.min.js:2:21280
    at jquery.min.js:4:29596
```

Which in boxesapp is at:

```
const listenersMap = /* @__PURE__ */ new WeakMap();
function addEventListener(ctx, type, listener, options) {
  let listeners;
  if (listener == null) {
    return;
  } else {
    const listeners1 = listenersMap.get(ctx);
    if (listeners1) {
      listeners = listeners1;
    } else {
      listeners = [];
      listenersMap.set(ctx, listeners); // * error here
    }
  }
```

So I'm wondering if crank can catch this error and fail silently if this is appropriate? Or am I barking at the wrong tree? I only post this here because it seems to me that this will pop up for others when jQuery is present.

I did spend quite a bit of time trying to adjust my code to eliminate the error but without success. I defer to better minds than my own.

Kind regards,
Darryl Cousins

[...]

PR #238, opened 2022-08-08
Fix eventtarget errors

brainkim:
Fixes https://github.com/bikeshaving/crank/issues/236

The `addEventListener`, `removeEventListener`, `dispatchEvent` seem to be leaked when using vite, causing runtime errors.


brainkim:
@darrylcousins If you can confirm that this fixes #236 that would be helpful! No worries otherwise, I think the code changes are fine anyways.

Issue #257, opened 2023-05-20
The type="text" attribute on input elements is being removed


waynebaylor:
Crank version: 0.5.3

If you render the following:

```jsx
import {renderer} from "@b9g/crank/dom";

function Form() {
  return (<form>
    <input type="text" />
    <input type="password" />
    <input type="email" />
  </form>);
}

renderer.render(<Form />, document.body);
```

The resulting html seen in the browser is:

```html
<form>
  <input>
  <input type="password">
  <input type="email">
</form>
```

The `type="text"` attribute is getting stripped off the first input. 

Without the `type="text"` attribute CSS rules that target `input[type="text"]` won't work.

brainkim:
INVESTIGATING

brainkim:
Okay my current working theory is that `type="text"` is the default, and the Crank DOM attribute/property algorithm is incorrectly flagging it as already set. @waynebaylor I am deeply sorry for any inconvenience this has caused 🙇‍♂️🙇‍♂️🙇‍♂️. As always thank you for the reproduction.

If my theory is correct, expect a fix in 0.5.4 within the hour. Otherwise, I will continue to investigate.

I’m slightly concussed from a bicycle accident right now but I am so glad you have attrited   yet, Wayne!

brainkim:
Okay. I figured out how to fix it but I am not going to, and I will explain why:

The default type of input elements is text (https://www.w3schools.com/tags/att_input_type.asp), so even if it looks wrong in a DOM inspector, it will actually behave as expected. You can double confirm by inspecting the element and logging `inputEl.type`.

Unless there are actual additional consequences which I am not aware of, this should be fine for the end user.

Nevertheless, if this kind of thing does bother you, I encourage you (or anyone) to make a PR! As a hint, the place to fix this would be in `/dom.js` under the method named “patch.”

Again, thank you for opening an issue!
BK

waynebaylor:
Sorry to hear about the accident, hope you're feeling better! 

I noticed this because a screen I'm working on has some CSS rules like:

```css
input[type=email],
input[type=number],
input[type=password],
input[type=search],
input[type=text] {
 border:1px var(--input-style) var(--font-color);
 width:100%;
 padding:.7em .5em;
 font-size:1em;
 font-family:var(--font-stack);
 -webkit-appearance:none;
 border-radius:0
}
```

and the CSS wasn't getting applied to bare `<input>` elements.

[...]

PR #258, opened 2023-05-21
Set node property even when equal to default prop value

waynebaylor:
Fix for #257 

Added a test to highlight the issue.

My fix might be heavy handed, so totally open to feedback.

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -162,7 +162,9 @@ export const impl: Partial<RendererImpl<Node, string>> = {
 						descriptor != null &&
 						(descriptor.writable === true || descriptor.set !== undefined)
 					) {
-						(node as any)[name] = value;
+						if ((node as any)[name] !== value || oldValue === undefined) {
+							(node as any)[name] = value;
+						}
 						return;
 					}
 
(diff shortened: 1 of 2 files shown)


brainkim:
Okay. I am awake and a little hung over. Thank you for this PR and sorry if this is blocking you. I am trying to stay away from the laptop so I wanted to demonstrate my appreciation for your help by providing my thoughts impromptu and hoping you can figure out the rest.

My initial thoughts are: This is a situation involving the initial render. This means that we're likely working with new DOM nodes which have no attributes or properties yet. So maybe we could do a check based on that? That conditional which you are trying to change does some stuff which I can't recall right now. Hot code. Live wire stuff.

By not setting properties when they are triple equal, ah yes now I remember, the main thing we are trying to stop is unexpected form input mutations? How do I say this plainly without looking up actual issues? Or maybe that's the opposite for this situation?

I guess, in a general sense, DOM nodes can change independently of rendering, so we're trying to treat it with kid gloves. This specific branch was likely put in place because it is important not to change attributes for some reason. The GitHub mobile view is not letting me see the full layers  of indentation.

TL;DR can we disable the branch for just the _initial_ render???

brainkim:
Also thank you for the test!

brainkim:
Perfecto! Thank you so much!

PR #266, opened 2023-10-11
Provide example of when CSS object is preferred over string

canadaduane:
It should have been obvious, but this bit me for a couple of days as I tried to figure out why my animations weren't working properly. Maybe someone else might benefit from the example in the future :)


brainkim:
Thanks!

PR #277, opened 2024-05-25
Update doc to mention context instead of this

tschaub:
This updates the website documentation to mention the context passed to components instead of `this` (as of #274).


brainkim:
Thanks!

Issue #281, opened 2025-04-17
Specify the attribute to assign a value to Attr or Prop


laamfun:
I use CrankJS and Ionic to develop an H5 application. Ionic uses the WebComponent specification. I need to follow the official documentation to pass in certain properties as Prop instead of Attr. I checked the documentation of CrankJS and found no relevant content.

[Here is the smallest display of the problem I am facing](https://stackblitz.com/edit/vitejs-vite-4keumyhk?file=src%2Fmain.jsx)

I was wondering, if I didn't miss anything, whether there can be the following syntax for such a requirement to be handled clearly, just like the preact processing method, can it be considered? For now, I need it very much.

```jsx
<div prop:demo1="demo1" attr:demo2="demo2"> test-div </div>
```

[Preact support for prop: & attr: directives](https://github.com/preactjs/preact/pull/4419)

brainkim:
There is no way to specifically update element properties or attributes. Crank will try to choose the best option. You can see the code for updating props/attributes here:

https://github.com/bikeshaving/crank/blob/master/src/dom.ts#L79-L199

Typically Crank will try to update element properties before defaulting to using attributes, unless attributes is better. What is the specific problem you’re having with properties/attributes?

brainkim:
If you want to set an attribute to `"false"` you can use the string `"false"`:

    <my-el attr="false" />

laamfun:
My needs:
Explicitly bind useHash to Prop

Problems encountered:
The following two situations are not bound correctly

```jsx
<ion-router useHash={false} />
<div useHash={false}>test</div>
```

My current treatment:
It is feasible to get Ref in the life cycle to set it manually

```jsx
async function* App() {
  let router;
  let div;

  for await ({} of this) {
    yield (
      <>
        <ion-router ref={(el) => (router = el)} />
        <div>test</div>
      </>
    );

    router.useHash = false;
    div.useHash = false;
  }
}
```

I think this is unreasonable. This is a common requirement in the use of WebComponent. The mainstream frameworks I understand seem to have relevant processing specifications, so I think it is possible to add functions to this in CrankJS. As for the specification, I recommend the PreactJS/SolidJS method.

[VueJS v-bind :_.prop / :_.attr](https://vuejs.org/api/built-in-directives.html#v-bind)
[PreactJS](https://github.com/preactjs/preact/pull/4419)
[SolidJS attr:\*](https://docs.solidjs.com/reference/jsx-attributes/attr#attr)
[...]

PR #282, opened 2025-04-22
implement prop: and attr: namespace for props

brainkim:
Implements #281


brainkim:
Thought a little bit about whether it’s wise to use an attribute namespace (`prop:`, `attr:`). It could potentially cause issues with exotic XML things, but I think it’s better than the alternatives. `prop$`, `attr$` is ugly and not as clear. `.` and `#` are invalid characters. `prop-` and `attr-` could potentially have similar issues. `prop:`/`attr:` is probably fine and only requires that JSX parsers don’t `throwIfNamespace`. Crank.js already doesn’t have a problem with namespaced attributes so this is a non-issue.

Issue #245, opened 2023-01-16
Async unmounting/updating


brainkim:
Tracking another attempt at https://github.com/bikeshaving/crank/pull/136 for 0.5.0

Issue #158, opened 2020-09-21
Stop blowing away the children of DOM nodes which we render into


brainkim:
Currently when we call `renderer.render(<App />, node)`, all of the children of `node` which Crank doesn’t own are blown away. This means that rendering into `document.body` will blow away the rest of the document. I would like Crank to work more cooperatively with the existing DOM; maybe we can keep track of the first and last nodes which we have rendered into a node and only render in that range.

cristianfalcone:
Will a fix for this allow an hydration render? Currently I'm having issues trying to identify when dom nodes were server side rendered.

https://github.com/cristianfalcone/pedal/issues/1

brainkim:
@cristianfalcone Can you elaborate a little on what you’re looking for? Perhaps you could add a `data-ssr` attribute to your routes for SSR code perhaps, which is erased or set to false on the client?

I still think that the heroic effort of reusing HTML-generated DOM nodes, and making sure they match between server and client during the initial handoff of the page is mostly unnecessary, as I’m quoted as saying in this issue (https://github.com/bikeshaving/crank/issues/34). Blowing away those nodes and creating new ones is fine for most use-cases, and I can think of workarounds for media elements and FOUC issues, so I haven’t been prioritizing it. Until someone creates a concrete demo of a UX issue which is caused by not reusing nodes and proves that there are no possible workarounds, I’m inclined to believe hydration is a waste of effort.

This specific issue is less about hydration and more just about being a good DOM citizen. I want to be able to call `renderer.render()` with `document.body` and not blow away everything else which was in the body element. I’ve been working with content editables more recently, which poses a similar issue, how should Crank behave when a rendered DOM is mutated by something besides the renderer. The common thread between all these issues is dealing with DOM nodes that are unknown to Crank, and I’d like to come up with a coherent philosophy at some point for how we deal with them, in a way that isn’t just blow up like React does.

Thanks for trying out Crank! I’ve been gearing up to create a similar sort of SSR solution but hit a hard wall when I came to the issue of bundling. I think I have a good grasp about what routing, navigation, and some of the other stuff should look like, but just haven’t really gotten to writing all of the code and bringing it together, and I’ve more or less been frustrated by everything from the current state of node/deno and their http servers, to the fact that no bundler seems to have a fleshed-out programmatic API. I’m happy to share my thoughts on routing or SSR or even bundling in a GitHub Discussion or by email. I’ll try to look through pedal soon!

brainkim:
[...]

Issue #68, opened 2020-04-26
async rendering of children: should be properly coordinated


ged-odoo:
Dear Author of Crank,
as requested in a reddit thread ([reddit ](https://www.reddit.com/r/javascript/comments/g4hia2/introducting_crank_modelling_the_component/fonndja?utm_source=share&utm_medium=web2x)), here is a small scenario that I think is not handled properly by crank:

```jsx
/** @jsx createElement */
import { createElement } from "@bikeshaving/crank";
import { renderer } from "@bikeshaving/crank/dom";

function wait(i) {
  return new Promise(resolve => {
    setTimeout(resolve, i);
  })
}
async function A({i}) {
  await wait(1000);
  return <div>A:{i}</div>;
}

async function B({i}) {
  await wait(200);
  return <div>B: {i}</div>;
}

function *Parent() {
  let i = 1;
  this.addEventListener('click', () => {
    i++;
    this.refresh();
  });
  while (true) {
    yield <div><A i={i}/><B i={i}/></div>;
  }
}
renderer.render(<Parent />, document.body);
```
We have here a parent with two sub components A and B. These sub components render themselves at a different speed. Whenever I click on the parent, some information is propagated to the full sub tree. 
What happens is that we see that B is updated first, and then A. I think that it is wrong. Both should be updated at the same moment. Otherwise, the UI is not in a consistent state.

brainkim:
The name for this type of bug would be “tearing,” and I think you might find some helpful info on tearing in the React repo or elsewhere as well.

[Here’s a sandbox with your code.](https://codesandbox.io/s/crank-issue-68-7xlh2) I guess, my response would be that this is that this is the correct behavior for those two components. Typically, you should lift your async/stateful logic into as high an ancestor if you can, and if you need to synchronize two components you should do so in a shared parent with `Promise.all`:

ged-odoo:
First, feel free to close this issue whenever you want.  I did not come here with the goal to troll or to annoy you. 

I understand your answer.  But I think that it is a consequence of a "weak" asynchronous model: you push the coordination logic to the application layer. Your answer give the responsability to the end developer of making it work. 

This means that:
1. I may need to rewrite my code to push some logic out of a component (where it makes sense) and into another (where it may not make sense)
2. it could also be impossible/very hard. For example, if one of these two components is a library.

Note that all standard frameworks currently have the same issue. They require the developer to write/organize the code around the framework's limitation.

[...]

PR #283, opened 2025-05-30
Version 0.7

brainkim:
- [x] Allow refresh() to be passed a callback.
- [x] Remove deprecated special props
- [x] Switch to a two-pass render #284 #68
  - [x] Fix hydration
  - [x] Move unmount logic to the commit phase
- [x] Revamp error handling #287
  - [x] Stop unhandled promise rejections
  - [x] Make sure the promise passed back into async gen for await of components reject.
- [x] Create and retain `Text` nodes via the `text()` method rather than via the `arrange()` method. #288 
  - [x] Refactor RenderAdapter methods
  - [x] Allow Text with ref
  - [x] Tests for Text element
- [x] Switch default async generator behavior to the same as `for...of` #291?
- [x] Async unmounting #245
  - [x] preserve index
- [ ] Stabilize internal Renderer API
  - [x] `RenderAdapter` `remove()` method
    - [x] Call during `unmount`
    - [x] pass info about whether the host node removal is “nested”
  - [x] Make patch() take the full props and oldProps.
  - [x] Pass host info as objects
  - [ ] Think really carefully about all the methods and their names.
- [x] Allow the copy prop to mark specific props for copying #292
  - [x] selective copying for props
  - [x] selective copying for children
- [x] Improved hydration #289
  - [x] Warn on hydration props mismatch
  - [x] boolean hydrate prop
  - [x] selective hydrate for props
  - [x] selective hydrate for children
- [x] Allow class to take objects (classnames, clsx) #294 
- [x] Async mounting/~~updating?~~ #290
  - [x] Allow schedule to return a promise to defer mounting
  - [x] ~~Allow schedule to return a promise to defer updating~~
- [x] Implement Suspense/SuspenseList/memo() and add them to an async.ts utility module
- [x] isExecuting/isUnmounted
- [x] ~~handle attributes with namespaces~~ not necessary in modern browsers
- [x] Rename `this.flush()` to `this.after()`
- [x] Don’t remove DOM nodes we don’t control. https://github.com/bikeshaving/crank/issues/158


brainkim:
I’m stuck working on some tricky promise rejections.

brainkim:
I'm stuck on error handling. Async generator components which are pulling will cause unhandled promise rejections even if they are handled by parent executions. For instance

~~~
async function *Child() {
  for await ({children} of this {
    throw new Error("this is handled");
  }
}

try {
  await renderer.render(<Child />, document.body);
} catch (err) {
  // renderer.render throws the error but an unhandled promise rejections still happens
}
~~~

brainkim (commented):
FUCK YEAH FUCK YEAH FUCK YEAH https://github.com/bikeshaving/crank/pull/283/commits/e60547d20276edacfa35c9cedfe834ff2f866f7a

brainkim:
I’ve made progress on async `schedule()` callbacks, but have decided to make them only work for the initial render of a component. This limitation can cover the `<SuspenseList />` use-case, and there’s still a bunch of open questions about how it should work on re-render. For instance, if the second set of children fulfills before the first, what happens to the first `schedule()` calls’s promises? Should they ignored? We would have to distinguish `schedule()` calls in this async generator component somehow, and determine how schedule() calls from different renders interact.
```js
async function *Component() {
  for await ({} of this) {
    this.schedule(() => deferUpdate());
    yield <Children1 />;
    this.schedule(() => deferUpdate());
    yield <Children2 />;
  }
}
```

On top of this, there’s the fact that components typically produce the same DOM nodes, and therefore async `schedule()` callbacks wouldn’t have any effect in that case anyways. Restricting async `schedule()` callbacks to the initial render of the component is not a design decision so much as reducing scope for this feature for my sanity. I slightly dislike that `schedule()` is inconsistent between first and subsequent updates. Nevertheless, the open design questions indicate it’s better to defer it til later.

PR #296, opened 2025-08-14
Why Crank?

spullara:
This was a relatively low effort change via Claude that I think would help if folks land on this readme. I was trying to figure out why I should care about Crank and this description has me caring about it and wanting to try it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,53 @@
 
 The fastest way to try Crank is via the [online playground](https://crank.js.org/playground). In addition, many of the code examples in these guides feature live previews.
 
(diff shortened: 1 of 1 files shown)


brainkim:
I like the copy! It’s punchy and I agree that the README.md could use some justification. The example needs to be updated to not use a while (true) loop, but I’m happy to merge and tweak. Thanks for stopping by!

brainkim:
I wrote an essay on why Crank isn’t reactive, if you’d like to check it out: https://crank.js.org/blog/why-be-reactive/

PR #314, opened 2025-10-15
Introduce eslint-plugin-crank

kristiandupont:
I've been playing with Crank again and one thing that bothered me was that I had forgotten some of the basics. Also, while LLM's are happy to write Crank code, they really lean towards a number of React-isms, which I had to tell them over and over.

So I wanted to create a plugin for ESLint. I noticed that you already have the package on NPM and in here, but it seems to never have gone anywhere, so perhaps you would like to use this as a start for an official version? It works reasonably well, though being an obvious first iteration, I am sure we will need to refine it.


packages/eslint-plugin-crank/eslint.config.js
```diff
@@ -0,0 +1,47 @@
+import js from "@eslint/js";
+import tseslint from "typescript-eslint";
+import crankPlugin from "./dist/index.js";
+
+export default tseslint.config(
+  js.configs.recommended,
+  ...tseslint.configs.recommended,
```
brainkim: I’m actually not a fan of the `typescript-eslint` recommended set of rules, last I checked (they might have changed). Can we allow users to bundle rules themselves?

packages/eslint-plugin-crank/package.json
```diff
@@ -1,10 +1,42 @@
...
+    "jsx"
+  ],
   "license": "MIT",
-  "dependencies": {
-    "eslint-plugin-react": "^7.32.2"
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "peerDependencies": {
+    "eslint": ">=9.0.0"
```
brainkim: I haven’t done the migration to ESLint 9.0 yet because it seemed like a pain in the butt. Can we support lower versions 👉👈? It’s fine if not, I’ll just upgrade I guess.

packages/eslint-plugin-crank/package.json
```diff
@@ -1,10 +1,42 @@
...
+    "eslint",
+    "eslintplugin",
+    "crank",
+    "jsx"
+  ],
   "license": "MIT",
-  "dependencies": {
-    "eslint-plugin-react": "^7.32.2"
+  "engines": {
+    "node": ">=18.0.0"
```
brainkim: Why does an ESLint plugin need an engines requirement?

packages/eslint-plugin-crank/test_debug.js
```diff
@@ -0,0 +1,35 @@
+import { ESLint } from "eslint";
```
brainkim: Is this a Claude Code artifact for debugging?

packages/eslint-plugin-crank/README.md
```diff
@@ -1,17 +1,436 @@
...
+  let seconds = 0;
+  while (true) {
+    // ❌ Can cause infinite loops and doesn't receive prop updates
+    yield <p>Multiplied value: {seconds * multiplier}</p>;
+  }
+}
+
+function* MultiProp({ title, count }) {
+  for ({ title } of this) {
+    // ❌ Using 'count' but not extracting it from context
```
brainkim: Can this be extracted into a separate `prop-destructuring-consistency` rule?

brainkim: I have a branch with a single commit which implements this rule:

https://github.com/bikeshaving/crank/commit/b9f0996bfbd1102a05f8b66f7b98e130db23bd50

packages/eslint-plugin-crank/README.md
```diff
@@ -1,17 +1,436 @@
...
+- `className` → `class`
+- `htmlFor` → `for`
+- `dangerouslySetInnerHTML={{ __html: value }}` → `innerHTML={value}`
+
+**Auto-fix:** ✅ This rule provides automatic fixes.
+
+**Why this matters:**
+- React uses non-standard prop names to avoid JavaScript reserved words
+- Crank uses standard HTML attributes directly
+- This is one of the most common sources of bugs when migrating from React
```
brainkim: Sorry to hear that! I think `className` and `htmlFor` should just work, and we might want to create a compatibility shim for `dangerouslySetInnerHTML` perhaps.

packages/eslint-plugin-crank/eslint.config.js
brainkim: Ahh wait, is this the eslint plugin for this package itself, never mind.

packages/eslint-plugin-crank/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+export const configs = {
+  recommended: {
+    plugins: ["crank"],
+    rules: {
+      "crank/prefer-refresh-callback": "error",
+      "crank/prefer-props-iterator": "error",
+      "crank/prefer-lowercase-event-props": "error",
+      "crank/no-react-props": "error",
+      "crank/no-yield-in-lifecycle-methods": "error",
+      "crank/require-cleanup-for-timers": "error",
```
brainkim: The old `eslint-plugin-crank` had the following rules:

```
				// A common Crank idiom is to destructure the context iterator values
				// with an empty pattern.
				//   for (const {} of this) {}
				// TODO: Write a rule that checks for empty patterns outside of this
				// context.
				"no-empty-pattern": 0,
				"react/jsx-uses-react": 2,
				"react/jsx-uses-vars": 2,
				"react/jsx-no-undef": 2,
				"react/jsx-no-duplicate-props": 2,
```

The `no-empty-pattern` rule is a must for recommended, and I don’t mind relying on `eslint-plugin-react`, unless there’s a better way perhaps.

packages/eslint-plugin-crank/README.md
```diff
@@ -1,17 +1,436 @@
...
+```
+
+**Why this matters:**
+
+- Lifecycle methods are designed for side effects, not for producing values
+- Using `yield` or returning values can cause unexpected behavior in the Crank lifecycle
+- The Crank documentation explicitly warns against this pattern
+- Helps prevent subtle bugs related to component lifecycle
+
+#### `crank/require-cleanup-for-timers`
```
brainkim: This rule requires the usage of `this.cleanup()` exclusively, but you can also put the cleanup after the props loop, or use `try`/`finally` (old code pattern):

```js
function *Timer() {
  let seconds = 0;
  const interval = setInterval(() => this.refresh(() => {
    seconds++;
  }), 1000);
  for ({} of this) {
    yield <div>{seconds}</div>;
  }

  clearInterval(interval);
}
```

```js
function *Timer() {
  let seconds = 0;
    seconds++;
    this.refresh();
  }, 1000);

  try {
    for ({} of this) {
      yield <div>{seconds}</div>;
    }
  } finally {
    clearInterval(interval);
  }
}
```

I use both the `cleanup()` method and after loop props in various places, so it would be nice if this was allowed.

packages/eslint-plugin-crank/src/rules/prefer-refresh-callback.ts
```diff
@@ -0,0 +1,222 @@
...
+      description:
+        "Prefer callback pattern for context.refresh() calls that are final actions in callbacks to avoid forgetting to call refresh",
+      category: "Best Practices",
+      recommended: true,
+    },
+    fixable: "code",
+    hasSuggestions: true,
+    schema: [],
+    messages: {
+      preferRefreshCallback:
```
brainkim: There are definitely going to be situations where you want the callback-less `refresh()` API, like when handling complex imperative workflows. I’m worried about false positives and auto-fixing, but this should definitely give new users a better sense of how to avoid forgetting to call `refresh()`.

brainkim (changes requested):
Thank you for taking the time to do this! I left some thoughts on the rules. For the most part, this seems like an improvement to the linting situation.

There are two important footguns that I’ve been meaning to lint away in Crank:

1. mismatched props destructuring:
```js
function* Component({children}) {
	for ({} of this) {
		yield children;
	}
}
```
2. mismatched props default value destructuring:
```js
function* Component({children="default"}) {
  for ({children} of this) {
    // The default parameter should match
    yield children;
  }
}
```

These footguns are unfortunate, but necessary as far as I can tell. It seems like you’ve already implemented this in the `crank-props-iterator` rule but it should be a separate rule.

Adding these rules would not be a blocker for merge. I have a commit where I implemented the props matching logic: https://github.com/bikeshaving/crank/commit/b9f0996bfbd1102a05f8b66f7b98e130db23bd50, but never got around to the default value matching as well.

The only other nits I have is that Crank development uses Bun instead of Node (it’s nice it gives you a free unit testing library), but it’s fine.

Overall excellent work and I’m eager to get this merged and shipped ASAP! Thank you so much Kristian!

kristiandupont:
Wow, that was a lot of fast feedback! Much appreciated! I'll address it all this evening or over the weekend if I don't have time. 
A bunch of your comments about package.json and the comments and docs are actually because of AI slop. I will clean that up.

packages/eslint-plugin-crank/README.md
```diff
@@ -1,17 +1,436 @@
...
+<button onclick={handleClick}>Click me</button>
+<input onchange={handleChange} />
+<div onmouseover={handleHover} />
+```
+
+**Auto-fix:** ✅ This rule provides automatic fixes.
+
+**Why this matters:**
+- Crank follows standard DOM conventions, not React's camelCase
+- Using React syntax will cause your event handlers not to work
```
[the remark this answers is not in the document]
kristiandupont: I see! So, should we remove this rule? Or remove it from recommendations?

packages/eslint-plugin-crank/src/rules/prefer-refresh-callback.ts
kristiandupont: Right. This is the rule that I wrote by far the most tests for. I am sure there are plenty of corner cases yet. I think my stance on linting is that it should generally get it at least 80% right to be valuable and then you just accept that there will be the occasional disable comment? I have to do this now and then with the React rules as well. 
Let me know if you want things changed here.

packages/eslint-plugin-crank/src/index.ts
brainkim: `"no-empty-pattern"` is a core ESLint rule, not a React one. You may be right about `react/jsx-uses-react`.

packages/eslint-plugin-crank/README.md
```diff
@@ -1,17 +1,436 @@
...
+rules: {
+  ...crankPlugin.configs['react-migration'].rules,
+}
+```
+
+## Rules
+
+### Core Rules
+
+#### `crank/no-yield-in-lifecycle-methods`
```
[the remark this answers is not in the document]
brainkim: I think as long as the rule doesn’t complain about `async`/explicit `return`, it’s fine and should be recommended.

packages/eslint-plugin-crank/eslint.config.js
brainkim: Yes that sounds good.

kristiandupont:
Sorry about disappearing here, something came up in my personal life. I won't have time to do any work on this right now, so if you want to create a branch that we can merge this into, you can change things around whichever way you like. Otherwise I will look into it when I have some time again.

PR #316, opened 2025-10-25
Set Default Text Color in Playground

brainkim:
Add inline styles to playground iframe to ensure text is visible in both dark and light modes. Previously, the text color would default to black before the external CSS loaded, making text unreadable on dark backgrounds.

Changes:
- Add proper <html> tag structure to match Python iframe implementation
- Add inline <style> block with CSS variables for immediate color application
- Set explicit text and background colors on body and all elements
- Ensures colors are applied before user code executes

### DIFF
--- a/website/src/utils/color-scheme.ts
+++ b/website/src/utils/color-scheme.ts
@@ -90,9 +90,8 @@ export function getColorSchemeScript(): string {
 export function syncIframes(scheme: ColorScheme): void {
 	if (typeof window === "undefined") return;
 
-	const iframes = document.querySelectorAll<HTMLIFrameElement>(
-		".playground-iframe",
-	);
+	const iframes =
+		document.querySelectorAll<HTMLIFrameElement>(".playground-iframe");
 
 	for (const iframe of iframes) {
 		// Send message to iframe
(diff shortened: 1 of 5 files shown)


website/src/components/root.ts
```diff
@@ -5,22 +5,12 @@ import {extractCritical} from "@emotion/server";
...
-			(window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches
-				? "dark" : "light"
-			);
-		if (colorScheme === "dark") {
-			document.body.classList.remove("color-scheme-light");
-		} else {
-			document.body.classList.add("color-scheme-light");
-		}
-	})()`;
+	const scriptText = `(() => { ${getColorSchemeScript()} })()`;
```
brainkim: The `getColorSchemeScript()` utility could be a component instead.

website/src/utils/color-scheme.ts
```diff
@@ -0,0 +1,127 @@
...
+
+		const isDark = colorScheme === "dark";
+		const bgColor = isDark ? "#0a0e1f" : "#e7f4f5";
+		const textColor = isDark ? "#f5f9ff" : "#0a0e1f";
+
+		document.documentElement.style.setProperty("--bg-color", bgColor);
+		document.documentElement.style.setProperty("--text-color", textColor);
+
+		if (!isDark) {
+			document.documentElement.classList.add("color-scheme-light");
```
brainkim: Is this a necessary mutation?

Issue #181, opened 2020-12-26
iframe onload called after every render


spiffytech:
I'd like to render an `<iframe>` with a loading indicator, which StackOverflow says should be done by watching for the `load` event on the `iframe` to remove the indicator. When I `refresh()` my component to remove the loading indicator, the `iframe`'s `onload` fires again, triggering another refresh, endlessly. This causes the iframe to infinitely reload the `src` page, hammering the `iframe`'d web server.

I've worked around this by using a variable to track if I need to break the refresh loop (which still loads the `iframe` content twice), but is there a reason the `iframe` fires the `onload` endlessly?

```tsx
function* Iframe(
  this: Context,
) {
  let loading = true;

  for (let props of this) {
    console.log("yielding iframe");
    yield (
      <section>
        {loading && <div>Loading...</div>}
        <iframe
          src="https://httpbin.org/html"
          onload={() => {loading = false; this.refresh()}}
        />
      </section>
    );
  }
}
```

spiffytech:
It looks like I can stop the double render by tracking the last `src` value, comparing it against the one from `props`, and using `<Copy />` and `crank-key` so the `iframe` doesn't get rerendered. Is this expected to be necessary?

brainkim:
>  Is this expected to be necessary?

Nope. It’s definitely an issue, and users shouldn’t be expected to use copies here. The problem is that Crank currently naively sets every property and attribute for each DOM node on render, and `src` for iframes seems to be one of those props which definitely should not be set every render if it hasn’t changed. I was hesitant to change this behavior because at some point I thought that reading and writing to the DOM frequently would be a performance issue. As far as I can tell, the effects of reading props and attributes before writing to them seems to be marginal, and mostly better for performance, because the DOM isn’t as smart as I thought it would be and setting properties like `contentEditable` with the same value will actually invalidate layout, for instance.

I have a working branch for performance which reads the DOM before attempting to set properties and attributes. I’ll try to timebox the work and get it into a 0.3.8 release by tomorrow. 

If you’re wondering why we don’t check the previous props instead of reading the DOM directly, it has to do with one of Crank’s philosophies, which is that we actually treat the DOM as the source of truth rather than the virtual tree. This means we can avoid having to define things like `defaultValue` for uncontrolled components, and it also makes Crank a better fit for a rich text editing backend. I can go more into that if you’re interested.

Sorry for the bug!

brainkim:
Fixed in 0.3.8.

[...]

PR #321, opened 2025-11-27
Fix URL property comparison for src and href attributes

brainkim1:
## Summary
- Fix unnecessary DOM updates when using relative URLs for `src` and `href` attributes
- The DOM property returns the resolved absolute URL (e.g., `http://localhost/path`), while the prop value may be relative (`/path`)
- Now we resolve the prop value using `new URL(value, element.baseURI)` before comparing, maintaining the "DOM as source of truth" philosophy

This completes the fix from #181 - the original fix compared `element.src !== value` to avoid unnecessary updates, but didn't account for URL resolution causing `element.src` to return an absolute URL while the prop is relative.

## Test plan
- [x] Added tests for relative `src` on iframe
- [x] Added tests for relative `href` on anchor
- [x] Added tests for absolute URLs
- [x] All 43 DOM tests pass
- [x] Linting passes

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -462,7 +462,24 @@ export const adapter: Partial<RenderAdapter<Node, string, Element>> = {
 						) &&
 						isWritableProperty(element, name)
 					) {
-						if ((element as any)[name] !== value || oldValue === undefined) {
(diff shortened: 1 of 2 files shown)


brainkim:
Thankee Claude

PR #329, opened 2026-01-22
Remove global document/window references from DOM renderer

brainkim:
## Summary

- Pass `root` element to adapter methods (`create`, `adopt`, `text`, `raw`) via the RenderAdapter interface
- Use `root.ownerDocument` to get document reference instead of relying on global `document`
- Use `element.ownerDocument.defaultView` instead of global `window` in hydration warnings

This enables using the DOM renderer in environments where `document` is not a global.

## Test plan

- [x] All existing tests pass (507/507)
- [x] TypeScript type checking passes
- [x] ESLint passes

### DIFF
--- a/src/crank.ts
+++ b/src/crank.ts
@@ -921,12 +921,11 @@ export class Renderer<
 	/**
 	 * @internal
 	 * FinalizationRegistry to automatically unmount when root nodes are garbage collected.
-	 * Uses WeakRef to avoid preventing the root from being collected.
 	 */
 	declare registry:
 		| FinalizationRegistry<{
 				adapter: RenderAdapter<TNode, TScope, TRoot, TResult>;
-				retRef: WeakRef<Retainer<TNode, TScope>>;
+				ret: Retainer<TNode, TScope>;
 		  }>
 		| undefined;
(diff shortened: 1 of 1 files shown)


src/crank.ts
```diff
@@ -1444,9 +1448,10 @@ function commit<TNode, TScope, TRoot extends TNode | undefined, TResult>(
 				el as Element<Text>,
 				scope,
 				hydrationNodes,
+				ctx?.root,
 			);
 		} else if (tag === Raw) {
-			value = commitRaw(adapter, host, ret, scope, hydrationNodes);
+			value = commitRaw(adapter, host, ret, scope, hydrationNodes, ctx?.root);
```
brainkim: no optional property syntax in this repo

src/dom.ts
```diff
@@ -105,10 +110,12 @@ export const adapter: Partial<RenderAdapter<Node, string, Element>> = {
 		tag,
 		tagName,
 		scope: xmlns,
+		root,
 	}: {
 		tag: string | symbol;
 		tagName: string;
 		scope: string | undefined;
+		root: Element | undefined;
```
brainkim: root is always available in the DOM renderer but I'm not sure TypeScript knows that

Issue #248, opened 2023-02-23
Better error messages for tagged template function Syntax Errors.


brainkim:
The 0.5 template tag needs better error messages.
``````js
jsx`
  <a
    href=${doc.url}
    aria-current=${doc.url === url && "page"}
    class=${css`
      text-decoration: none;
    `}
    "
  >${doc.attributes.title}</a>
`
``````
Throws

```
file:///Users/briankim/Projects/crank/src/jsx-tag.ts:248
                                                        throw new SyntaxError(
             ^
SyntaxError: Unexpected text `"`
```

These messages could be nicer. Like how Babel and TypeScript do it. Something that shows the error in context of the line. Maybe we could use a `SyntaxError` subclass like `JSXSyntaxError` to make things even clearer.

Issue #90, opened 2020-05-02
React compatibility


dfabulich:
It would be cool and useful if Crank were to provide a compatibility layer to consume React components. This would provide a migration path from React to Crank, porting components from the top down.

Alternately (or as well), it could be possible to wrap Crank components as React components, allowing users to do a bottom-up migration from React to Crank.

brainkim:
This is not something I’m personally prioritizing at the moment because I imagine it to be very difficult. Also, I think there is a lot of unlearning people will have to do with regards to hooks, not doing side-effects in the render method, and opting into updates rather than opting out of them. As far as wrapping Crank components to use in React, I think the best way to do this is, ironically web components. The JSX typings would clash if we just tried to import Crank into a React application, and there would be no benefit of sharing elements between the two libraries, insofar as elements mean different things.

If someone is interested in this sort of stuff, please do experiment! I’m just trying to focus on other stuff right now and would only consider this once Crank’s ideas and design patterns are clearer.

dfabulich:
> As far as wrapping Crank components to use in React, I think the best way to do this is, ironically web components.

I did think about that, but IMO React doesn't make it easy to consume web components, especially when passing props that aren't just strings. It's easy to set string props (as attributes on the custom element), but anything else is kind of a mess. 

https://reactjs.org/docs/web-components.html "If you are using third-party Web Components, the best solution is to write a React component that behaves as a wrapper for your Web Component." 

So maybe this has to wait until Crank has good web component support (see issue #47), but I think whatever work is required to create a WC generator, it wouldn't be much _more_ work to create a React-component wrapper for the generated WCs.

As for consuming React components in Crank, I didn't mean "Crank itself should be able to directly consume React," but just that if you run crank.js and react.js in the same page, (or maybe just preact.js?) it should be possible to put a React component in a Crank app. I guess WCs could be used for that, too… but again you've got the attroperties problem

brainkim:
> So maybe this has to wait until Crank has good web component support (see issue #47), but I think whatever work is required to create a WC generator, it wouldn't be much more work to create a React-component wrapper for the generated WCs.

Agreed

> As for consuming React components in Crank, I didn't mean "Crank itself should be able to directly consume React," but just that if you run crank.js and react.js in the same page, (or maybe just preact.js?) it should be possible to put a React component in a Crank app. I guess WCs could be used for that, too… but again you've got the attroperties problem

[...]

PR #339, opened 2026-02-15
Release 0.7.6

### LINKED ISSUE #256: Export `renderer` and `Renderer` based on environment from `@b9g/crank/standalone`
It’s annoying to import the new jsx template tag from `standalone`, and also have to import a renderer for the respective environment. I’m wondering if we can use some sort of environment detection mechanism to have the correct one imported.

Before:

```js
import {jsx} from "@b9g/crank/standalone";
import {renderer} from "@b9g/crank/dom";

renderer.render(jsx`<div id="hello">Hello world</div>`, document.body);
```

After:
```js
import {jsx, renderer} from "@b9g/crank/standalone";

renderer.render(jsx`<div id="hello">Hello world</div>`, document.body);
```

How do we determine environment? I don’t like using non-standard package.json exports, and exports conditions are really confusing. I honestly just want to do top-level await imports.

brainkim:
## Summary

Closes #256. Related to #90, #248, #341.

### New features
- **Standalone re-exports renderers** — `@b9g/crank/standalone` now exports `renderer`, `domRenderer`, `DOMRenderer`, `htmlRenderer`, and `HTMLRenderer` for single-import CDN/no-build usage
- **React prop compatibility** — Both renderers accept `htmlFor`, `className`, `dangerouslySetInnerHTML`, and SVG camelCase attributes so copy-pasted React JSX works without modification. When both the React alias and the native prop are present (e.g. `htmlFor` and `for`), the native prop takes precedence. SVG prop mapping only applies within SVG elements.
- **JSX template tag error context** (#248) — Syntax errors now include line/column context with a caret pointer. Error parses are no longer cached.

### Bug fixes
- **Fix `dangerouslySetInnerHTML` rendering literal "null"/"undefined"** — Both renderers now require the `{__html: "..."}` shape and ignore other value types
- **Fix `attr:` prefix re-adding removed attributes** (#341) — `null`/`false` values now correctly remove the attribute instead of setting it to `"null"`/`"false"`
- **Fix `foreignObject` namespace handling** — Children are created in HTML namespace without requiring an explicit `xmlns` prop; `foreignObject` itself retains SVG namespace and prop normalization
- **Fix Portal not reading root namespace** — Portal reads `root.namespaceURI` so portaling into an SVG element creates SVG children, and portaling out of SVG resets to HTML namespace

### Internal improvements
- **Reduce GC churn in DOM patch and child diff** — Eliminated spread merges in `patch()`, added single-child `diffChild()` fast path, replaced `wrap()` with inline iteration

## Test plan
- [x] `bun run typecheck` passes
- [x] `bun run test` passes — 563 core + 126 eslint plugin tests
- [x] `bun run build` succeeds
- [x] CI green across chromium, firefox, webkit
- [x] New tests cover: htmlFor, className, dangerouslySetInnerHTML, SVG attrs, attr: prefix removal, foreignObject namespace, Portal scope, prop precedence, prop patching, child cardinality transitions, JSX tag error context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,59 +9,58 @@
   ```
 
 - **React prop compatibility**
-  The DOM and HTML renderers now silently handle React-style props, so
-  copy-pasted React JSX works without modification:
+  The DOM and HTML renderers now accept React-style props, so copy-pasted React
+  JSX works without modification:
   - `htmlFor` → sets the `for` attribute
+  - `className` → sets the `class` attribute
   - `dangerouslySetInnerHTML={{__html: "..."}}` → sets `innerHTML`
   - SVG camelCase attributes (`strokeWidth`, `fillOpacity`, `textAnchor`, etc.)
-    → converted to standard kebab-case (`stroke-width`, `fill-opacity`,
-    `text-anchor`)
+    → converted to kebab-case (`stroke-width`, `fill-opacity`, `text-anchor`)
 
-  The ESLint plugin continues to guide users toward idiomatic Crank prop names.
+  When both the React alias and the native prop are present (e.g. `htmlFor` and
+  `for`), the native prop takes precedence. SVG prop mapping only applies within
+  SVG elements, so HTML props like `tabIndex` are unaffected.
 
-- **Better JSX template tag error messages**
-  Syntax errors in `jsx` template literals now include visual context with a
-  caret pointer:
+  The ESLint plugin guides users toward idiomatic Crank prop names.
+
+- **JSX template tag error context**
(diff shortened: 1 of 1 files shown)


src/_svg.ts
```diff
@@ -0,0 +1,69 @@
+// React SVG camelCase → standard SVG attribute mapping.
+// Only includes entries where the React name differs from the SVG attribute name.
+// Source: packages/eslint-plugin-crank/src/rules/no-react-svg-props.ts
```
brainkim: @claude Can we find a canonical source for this insanity?

PR #348, opened 2026-02-26
Add User Timing API instrumentation

brainkim:
## Summary

- Adds `performance.mark`/`performance.measure` instrumentation so component render times show up in the browser Performance panel's Timings lane
- Instruments component execution in `runComponent` (sync/async functions, sync/async generators) and render phases (`⚙ diff`, `⚙ commit`) in `renderRoot`
- Always-on — no flag, no separate import, no dev/prod split
- Guarded with `typeof performance !== "undefined"` so the HTML renderer (SSR) isn't affected

## What it looks like in DevTools

Record a Performance trace → Timings lane shows:
```
⚙ diff           |████████████████|
⚙ MyComponent        |████|
⚙ ChildComponent          |██|
⚙ commit                           |██████|
```

## Test plan

- [x] `bun run build` passes
- [x] `bun run test` passes (573 tests)
- [x] `bun run lint` passes
- [x] Manual: open a Crank app, record Performance trace, verify marks in Timings lane

### DIFF
--- a/website/bun.lock
+++ b/website/bun.lock
@@ -11,7 +11,7 @@
         "@b9g/platform-bun": "^0.1.16",
         "@b9g/revise": "^0.1.3",
         "@b9g/router": "^0.2.5",
-        "@b9g/shovel": "^0.2.8",
+        "@b9g/shovel": "^0.2.16",
         "@emotion/css": "^11.13.5",
         "@emotion/server": "^11.11.0",
         "@lezer/javascript": "^1.5.4",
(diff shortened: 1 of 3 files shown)


src/crank.ts
```diff
@@ -15,6 +15,27 @@ import {
...
+		const name = "⚙ " + label;
+		try {
+			performance.measure(name, name);
+		} catch (_) {
+			// Mark may not exist
+		}
+		performance.clearMarks(name);
+	}
+}
+
```
brainkim: These functions may be better placed in src/_utils.ts

brainkim:
Screenshot from the Crank Homepage Gear hero interactive:

Screenshot 2026-02-26 at 2 04 52 AM

Seems pretty nice and straightforward. Still thinking if we should support the other timing API which React eventually did use.

PR #357, opened 2026-04-02
Don't mutate caller's props object in createElement

### LINKED ISSUE #356: Bug when using Crank.js with vanity-h
I want to use **Crank.js** with **vanity-h**, a library I developed. I did not encounter the following issue when integrating it with Vue previously.

vanity-h: https://github.com/VanityH/vanityh
Core logic of vanity-h: https://github.com/VanityH/vanityh/blob/main/src/index.ts
Vue demo: https://stackblitz.com/edit/vitejs-vite-bujruupx
Crank.js bug demo: https://stackblitz.com/edit/vitejs-vite-qqikysvw

Vanity-h is a library developed by me. Understanding this issue may take some time, but I will do my best to explain it clearly.

The specific problem is that when I use `yield div("divA", div("divB"));`, the console throws an `Uncaught RangeError: Maximum call stack size exceeded` error (from crank-C3aCAsya.js?v=2558a0a4:735).

However, the code runs normally when I use `div.class('class-name')("divA", div("divB"));`.

This may not be a bug in Crank.js, but after trying, I still haven't been able to identify the root cause of the issue. Sorry for the trouble.

brainkim:
createElement was mutating the incoming props to assign children, which causes circular references when callers reuse the same props object across calls (e.g. vanity-h's proxy-based hyperscript DSL). Spread into a fresh object instead, only when children are present.

Also guard the key assignment in jsxAdapter behind a null check to avoid unconditionally adding key: undefined to every props object, which causes unnecessary hidden class transitions.

Fixes #356

### DIFF
--- a/src/crank.ts
+++ b/src/crank.ts
@@ -270,9 +270,9 @@ export function createElement<TTag extends Tag>(
 	}
 
 	if (children.length > 1) {
-		(props as TagProps<TTag>).children = children;
+		props = {...props as TagProps<TTag>, children};
 	} else if (children.length === 1) {
-		(props as TagProps<TTag>).children = children[0];
+		props = {...props as TagProps<TTag>, children: children[0]};
 	}
 
 	return new Element(tag, props as TagProps<TTag>);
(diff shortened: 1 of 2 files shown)


brainkim:
Superseded by #366, which folds in the same `createElement` props-mutation fix (identical `{...props, children}` spread and `jsx-runtime` `key` guard) plus test coverage, on top of the 0.7 deprecation-shim removal. Closing in favor of that.

Issue #60, opened 2020-04-24
Strongly typed intrinsic elements


brainkim:

wmadden:
There's now discussion on a few separate issues about what things are worth typing and which end-user scenarios should be covered by typing or not. It'd be great to have a single, clear written description of what the Crank API goals are, what motivations to consider for strict typing etc - a manifesto or something similar.

Some goals are clear: use JavaScript language features instead of building framework solutions, and don't copy existing frameworks' design decisions just out of familiarity. What motivates API design decisions beyond that? 

E.g. regarding types I think it's important to strongly type the `onevent` props, since it's the base that people will build everything else on. More generally, I think whenever a user is writing JSX, props for the components they're using should be strongly typed, and the inputs and outputs of their component functions should be strongly typed.

I'm sure not everyone will agree with me, but it'd be good to have a clear direction in one place to reference for these discussions.

PR #362, opened 2026-06-13
@b9g/jsx-web-types — typed JSX intrinsic elements (#60)

brainkim:
Introduces `@b9g/jsx-web-types` and wires it into Crank's `JSX.IntrinsicElements` (#60). Spec-driven, framework-agnostic types for HTML, SVG, and MathML intrinsic elements.

## What it is
- A generated `.d.ts` package: element → attributes/properties, ARIA, microdata/RDFa, global attributes, and a DOM event map. Each attribute carries its MDN description.
- The standard types use native platform names (`class`, `onclick`). A separate **camelCase variant** (`@b9g/jsx-web-types/camel-case`) re-keys them for React/Inferno (`className`, `htmlFor`, …), with names sourced from React's own `possibleStandardNames.js`.
- Crank's `JSX.IntrinsicElements extends WebIntrinsicElements<…>`, supplying its own intrinsic attributes (`key`/`ref`/`prop:`/`attr:`), children attribute, and object-`style`/`class` value overrides via the `TPropOverrides` parameter — so Crank's conventions live in the parameters, not the base types.

## Everything is sourced — no hand-coded platform data
Generation is a reproducible function of authoritative sources:

| Source | What it provides | How accessed |
|---|---|---|
| **BCD** (`@mdn/browser-compat-data`) | element names, per-element attributes, `global_attributes`, deprecated/experimental status, `mdn_url` | imported JSON |
| **MDN pages** | human descriptions + code examples (the hover text) | scraped with `cheerio`, cached in `data/mdn-cache.json` (offline-capable; network only on a cache miss) |
| **`aria-data`** | ARIA attributes and their value enums | imported JSON |
| **TypeScript `lib.dom.d.ts`** | tag → DOM interface maps, DOM property types, the `*EventMap` interfaces | parsed with `ts-morph` |
| **W3C RDFa Core spec** | RDFa attributes | `scripts/parse-rdfa-w3c.ts` → `data/rdfa-attributes.json` |
| **MDN Microdata guide** | microdata attributes | `scripts/parse-microdata-mdn.ts` → `data/microdata-attributes.json` |
| **React `possibleStandardNames.js`** | the camelCase prop-name map (pinned to a release) | `scripts/parse-camel-case-names.ts` → `src/camel-case.generated.d.ts` |
| _(hand-written)_ | only structural constants — the namespace names and the fixed `XMLAttributes` W3C set | inline in the generator |

Every `.d.ts` and `data/*.json` is regenerable from these and stamped do-not-edit; there is no hand-frozen platform data.

> One known wrinkle: BCD omits `fill` and `systemLanguage` from `svg.global_attributes` despite their families being present. The generator promotes exactly those two (cross-referenced against BCD's per-element data); reported upstream as [mdn/browser-compat-data#29912](https://github.com/mdn/browser-compat-data/issues/29912), and the workaround deletes itself once that lands.

## Descriptions surface in editors (LSP)
Each attribute's MDN JSDoc reaches hover/completion. The prop types are composed so TypeScript preserves per-attribute doc comments (homomorphic mapped types; multi-namespace tags resolved by precedence rather than unioned), verified end-to-end through Crank JSX with the TypeScript language service — `<div class>`, `<a href>`, `<label for>`, etc. all show their descriptions.

## Children
Void elements aren't specially typed: every element accepts children (the runtime ignores stray ones), so there's no void set or `TIsVoid` plumbing to maintain.

## Tests
`bun run test` typechecks both `src/` and the type-level test suite in `tests/`. Root `tsc --noEmit` is clean.


packages/jsx-web-types/scripts/generate.ts
```diff
@@ -0,0 +1,1006 @@
...
+import prettier from "prettier";
+import {createRequire} from "module";
+
+const nodeRequire = createRequire(import.meta.url);
+
+// Locate the real TypeScript install's lib.dom.d.ts rather than assuming a
+// node_modules layout. require.resolve can land in a package store that omits
+// the bundled lib files (e.g. Bun's .bun store), so we also walk up from this
+// script looking for node_modules/typescript/lib, which handles hoisting.
+function resolveDomLib(): string {
```
brainkim: @claude Please use acrocase: https://acrocase.org

packages/jsx-web-types/README.md
```diff
@@ -0,0 +1,145 @@
...
+- ✅ **Smart unions**: properties like `width` accept both `number` and `string`
+- ✅ **Documented on hover**: each attribute carries its MDN description, surfaced in editor hover/completion
+- ✅ **ARIA support**: full ARIA attribute typing with value enums
+- ✅ **Framework agnostic**: native names for any JSX framework, plus a camelCase variant for React-likes
+- ✅ **Specification-driven**: generated from BCD/MDN, the WAI-ARIA spec, and TypeScript's DOM types
+
+## Installation
+
+```bash
+npm install @b9g/jsx-web-types
```
brainkim, line 17: Still debating if we should just go bare `jsx-web-types`

src/crank.ts
```diff
@@ -3842,8 +3847,33 @@ declare global {
 		// some reason.
 		// interface Element extends CrankElement {}
 
-		export interface IntrinsicElements {
-			[tag: string]: any;
+		// Crank-specific prop value conventions, layered onto the platform types
```
brainkim, line 3850: This is pretty cool. The design makes me feel like a cat bathing in the sun.

test/copy-prop.tsx
```diff
@@ -490,6 +490,7 @@ test("copy prop can be used for uncontrolled input values", () => {
 						type="text"
 						copy={force ? false : "value"}
 						value={value}
+						// @ts-expect-error - handler types the event as InputEvent; the native oninput prop uses the base Event
```
brainkim, line 493: why though?

packages/jsx-web-types/scripts/generate.ts
brainkim: Done in 2e1b3a0b — `resolveDomLib`→`resolveDOMLib`, and swept the rest of the script for consistency: `mdnUrl`→`mdnURL`, `isFragmentUrl`→`isFragmentURL`, `formattedHtml`→`formattedHTML`. The `mdnURL` locals now match the object keys they feed.

test/copy-prop.tsx
brainkim: The handler prop derives its event type from lib.dom’s `GlobalEventHandlersEventMap`, and lib.dom types `oninput` as the base `Event`, not `InputEvent` (long-standing quirk — the runtime fires an `InputEvent`, but the static type is widened). So a callback annotated `(ev: InputEvent)` is not assignable to the `(ev: Event) => …` the prop expects, hence the `@ts-expect-error`. The test pins that gap so we notice if the generated types ever start narrowing it.

packages/jsx-web-types/README.md
brainkim: Trade-off: bare `jsx-web-types` reads cleaner and is more discoverable for non-b9g consumers, but it means owning an unscoped name and it drops out of the `@b9g/*` family grouping (jsx-runtime, libuild, crankdown are all scoped). I lean scoped for consistency unless you want it to explicitly stand alone as a framework-agnostic package. Either way it’s a one-line rename — your call.

Issue #360, opened 2026-06-13
RFC: @b9g/crank/web-components — class-based custom elements


brainkim:
# `@b9g/crank/web-components`

- **Status:** Draft / for discussion
- **Target:** 0.8.0
- **Related:** #47, #318, #324, #60

## Summary

`@b9g/crank/web-components` adds `CrankHTMLElement`, a base class for Web Components written with
Crank. Subclass it, declare config as **static fields**, and define a `render` **method**.
`render` is a normal Crank component (same forms, same lifecycle), except `this` is the element
and `ctx` is an argument. No config object, no constructor.

```js
import {CrankHTMLElement} from "@b9g/crank/web-components";

class XBlinkElement extends CrankHTMLElement {
  static shadowDOM = true;

  *render(props, ctx) {
    let on = true;
    const id = setInterval(() => { on = !on; ctx.refresh(); }, 750);
    ctx.cleanup(() => clearInterval(id));
    for ({} of ctx) {
      yield <span style={`visibility: ${on ? "visible" : "hidden"}`}>
        <slot />
      </span>;
    }
  }
}

customElements.define("x-blink", XBlinkElement);
```

## The class

Config is static fields; behavior is methods and accessors.

| Static field | Type | Meaning |
|---|---|---|
| `observedAttributes` | `string[]` | Attributes that trigger a re-render. Read by the platform. |
| `events` | `readonly string[]` | Emitted event types; generates the `on<type>` handler properties. |
| `formAssociated` | `boolean` | `true` opts the element into form association (read by the platform). See Form association. |
| `shadowDOM` | `boolean` \| `ShadowRootInit` | Omitted/`false`: light DOM (default). `true`: open shadow root. A `ShadowRootInit`: that exact `attachShadow` config. |
| `styles` | `string` \| `CSSStyleSheet` \| `Array<string \| CSSStyleSheet>` | CSS, applied once per class via `adoptedStyleSheets`. See Styling. |

**Light DOM is the default.** Opt into shadow DOM with `shadowDOM = true` for encapsulation or
slots (the opposite of Lit's shadow-by-default).

Properties and the imperative API are ordinary class members. Config is class-level, so it lives
in statics rather than constructor arguments, and no constructor is needed.

The base adds one instance method, `requestUpdate()`: a microtask-batched re-render request. The
context is never exposed on the element; it reaches `render` as an argument and is otherwise
module-private.

## `render`

[...]

PR #363, opened 2026-06-22
Add @b9g/crank/web-components (CrankHTMLElement)

### LINKED ISSUE #47: Web Components: Discussion about how to use Crank.js for custom elements
Wouldn't it be great to implement custom elements in a "crankish" 😄 way?
This is a discussion thread to gather all ideas to be found about the question how to use Crank.js or Crank.js patterns to implement custom elements.

This is a brainstorming, nobody expects a fully sophisticated proposal. So please share every idea that comes to your mind: Requirements, API suggestions, best practices , dos and don'ts, known pitfalls etc.

brainkim:
Implements the class-based `CrankHTMLElement` from the RFC in #360 (closes #47).

## What's implemented
- Static config: `observedAttributes`, `events`, `formAssociated`, `shadowDOM`, `styles`.
- `render` method (`this` = element, `ctx` = arg), all four component forms via a stable bridge.
- `requestUpdate()` — microtask-batched, coalescing, safe before first connect.
- Light DOM (default, render owns the subtree) or shadow DOM with native `<slot>`.
- Faithful `on<type>` handlers (lazy activate/deactivate; reassignment preserves listener order), **typed from `CrankHTMLElement<{events: typeof Class.events}>`** (names-only for a tuple, typed payloads for an event map).
- `static styles` derived once per class (string / `CSSStyleSheet` / mixed array), adopted into the shadow root or `document`.
- Form association: base re-renders on the four form callbacks; `ElementInternals` stays the author's.
- Pre-upgrade properties re-applied through their accessors on connect.
- Render errors → cancelable `error` event, then `reportError` if unhandled.
- DOM renderer widened to accept a `ShadowRoot` (or document fragment) as a render root.

## Examples — `examples/web-components.tsx`
The RFC's Scope examples: `x-blink` (generator + timer, light DOM), `x-marquee` (shadow DOM, scoped `static styles`, imperative `start()`/`stop()`, typed `static events`), and `x-rating` (form-associated, `ElementInternals` value/validity, super-able reset/restore), plus a small demo wiring them together.

## Tests — `test/web-components.tsx`, 33 passing in Chromium
**Behavior:** sync/async/generator/async-generator render; attribute re-render + same-value bail; `requestUpdate` + coalescing; generator-state persistence; shadow + slot projection; open `ShadowRootInit` and closed mode; `on<type>` firing + listener ordering; in-component error fallback; render-error event (sync + async) + `reportError`; form FormData/reset/`:disabled`; light + shadow style adoption (deduped, mixed arrays); pre-upgrade property upgrade; light-DOM child replacement; move-vs-removal teardown.
**Integration:** rendering a custom element through Crank's own DOM renderer; nested custom elements; `ctx.refresh`/`schedule`/`after`; provisions within an element; computed shadow styles; accessor→attribute reflection. Plus a compile-only check of the event typing (tuple and event-map forms).
Full suite green (606 passing); `tsc --noEmit` clean; build emits the `./web-components` export (`.js`/`.cjs`/`.d.ts`).

Bugs the tests caught and this PR fixes: light DOM wasn't replacing authored children; a pre-connect `requestUpdate` fired a spurious extra render; pre-upgrade properties weren't upgraded.

## Out of scope (deferred by the RFC)
SSR / declarative shadow DOM / hydration (RFC Appendix), the form-value SSR shim, customized built-ins, and a scoped `css` prop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,16 @@
     "./umd.js": {
       "require": "./dist/umd.js"
     },
+    "./web-components": {
+      "types": "./dist/web-components.d.ts",
+      "import": "./dist/web-components.js",
+      "require": "./dist/web-components.cjs"
+    },
+    "./web-components.js": {
+      "types": "./dist/web-components.d.ts",
+      "import": "./dist/web-components.js",
+      "require": "./dist/web-components.cjs"
+    },
     "./package.json": "./dist/package.json"
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,22 @@
   and as a real line break in a `<pre>`. Whitespace adjacent to an element, an
   expression, or the template edge is still stripped as layout.
 
+### New API
+- **`@b9g/crank/web-components` — write Web Components with Crank**
+  A new subpath export with a `CrankHTMLElement` base class. Subclass it, declare
+  configuration as static fields (`observedAttributes`, `events`, `formAssociated`,
+  `shadowDOM`, `styles`), and write a `render` method that is a normal Crank
+  component: `this` is the element, `ctx` is an argument, and any of the four
+  component forms (sync, async, generator, async generator) is allowed. Light DOM
+  is the default; opt into shadow DOM with `static shadowDOM = true`. `static
+  events` generates typed `on<type>` handler properties, and `static styles` is
+  adopted once per class via `adoptedStyleSheets`. Form-associated elements get
+  re-rendering defaults for the four form callbacks.
+
+- **The DOM renderer now accepts a `ShadowRoot` as a render root**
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -110,6 +110,14 @@
     },
     "./umd.js": {
       "require": "./dist/umd.js"
+    },
+    "./web-components": {
```
brainkim, line 161: thoughts on shortening this to `@b9g/crank/wc`?

brainkim: Every other crank subpath is a spelled-out word (`/standalone`, `/dom`, `/html`, `/jsx-runtime`), so `/web-components` matches the family; `/wc` would be the lone abbreviation and less greppable. I lean `/web-components`, but the shorter import is a trivial change if you prefer it.

PR #364, opened 2026-06-23
Preserve the break between adjacent text lines in the jsx template (#359)

### LINKED ISSUE #359: jsx template strips the space between adjacent text lines (diverges from JSX whitespace rule)
The `jsx` tagged-template collapses a newline **between two text runs** to nothing, where standard JSX collapses it to a single space. So hand-written multi-line prose loses the spaces at its line breaks.

## Repro

```js
import {jsx} from "@b9g/crank/standalone";
import {renderer} from "@b9g/crank/html";

await renderer.render(jsx`<p>alpha
beta</p>`);
// actual:   "<p>alphabeta</p>"
// expected: "<p>alpha beta</p>"
```

In practice this shows up as joined words in any multi-line copy — e.g. a homepage paragraph rendering `promise-based APIfor creating async iterators`.

## How it compares to JSX

Babel/React's whitespace rule collapses an interior newline *within a text run* to a single space, but strips whitespace *adjacent to an element*. Crank matches JSX for the element-adjacent cases and diverges only for text–text:

| source | Crank | JSX (esbuild + tsc) |
|---|---|---|
| `alpha⏎beta` (text–text) | `alphabeta` ❌ | `alpha beta` |
| `alpha <b/>⏎beta` (elem–text) | `alpha <b/>beta` ✅ | `alpha <b/>beta` |
| `alpha⏎<b/>` (text–elem) | `alpha<b/>` ✅ | `alpha<b/>` |

Confirmed with two independent transpilers — `<p>alpha⏎beta</p>` compiles to `createElement("p", null, "alpha beta")` under both `esbuild --loader=jsx` and `tsc --jsx react`.

## Root cause

In `src/jsx-tag.ts`, the children tokenizer captures `(\r|\n|\r\n)\s*` as a token; the handler trims the trailing whitespace off the preceding text run, pushes it, then consumes the newline token **without emitting anything in its place**. So the text runs on either side land adjacent with no separator. It treats a text–text newline identically to an element-adjacent one, but JSX treats those two differently.

## Suggested direction

Collapse a text–text newline to a single space while preserving element-adjacent stripping: when a newline follows non-empty text, emit a single space if the next token is text, and drop it if the next token is a tag. That changes only the text–text case and leaves the (JSX-faithful, `{" "}`-requiring) element-adjacent behavior intact. Worth confirming against the existing whitespace tests to see whether anything relies on the current stripping.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

brainkim:
Fixes #359.

## The bug
The `jsx` tagged template dropped the whitespace **between two text runs** entirely, so a line break in multi-line prose vanished:

```js
jsx`<p>alpha
beta</p>`
// before: "<p>alphabeta</p>"
// after:  "<p>alpha\nbeta</p>"
```

## The fix
Whitespace between two text runs is now **preserved verbatim** — rather than dropped (the bug) or collapsed to a space (standard JSX). In `src/jsx-tag.ts` the children tokenizer, on a newline that follows non-empty text whose next significant character is also text (not `<`, and not the end of a span where an expression follows), appends the whitespace run to the preceding text as-is instead of discarding it. Only whitespace flanked by a non-text boundary — an element, an expression, or the template edge — is stripped.

## Why preserve, and not match JSX?
Standard JSX collapses a text–text newline to a space. That rule exists to make source **safe to reformat**: a formatter (Prettier, tsc) rewraps JSX text to a print width, so the newline you typed can't be significant — otherwise every reflow would change the output.

That rationale doesn't apply here. **Prettier has no embedded formatter for the `jsx` tag** — it leaves the template interior byte-for-byte. Nothing reflows the template, so a line break in it is the author's deliberate whitespace, not disposable layout. Preserving it is both more faithful and more expressive:

- It renders as a space in normal flow (the browser collapses it) and as a **real line break inside `<pre>`**, where collapsing to a space would be wrong. Because the indentation between text is kept too, a `<pre>`'s own formatting now survives without reaching for `${...}`.
- Anything still not expressible literally (e.g. whitespace hugging the `<pre>` tags) stays reachable through `${...}` interpolation, which bypasses whitespace handling entirely — the same escape hatch JSX offers with `{"\n"}`.

So this deliberately **diverges from JSX** for the tagged-template context. (The `html` alias shares this function; note Prettier *does* reflow `` html`` `` templates as HTML — a separate, pre-existing mismatch not addressed here.)

## Whitespace behavior

| source | result |
|---|---|
| `alpha⏎beta` (text–text) | `"alpha\n"`, `"beta"` |
| `alpha⏎⏎   beta` (blank line + indent) | `"alpha\n\n   "`, `"beta"` (verbatim) |
| `a   b` (interior run) | `"a   b"` (preserved) |
| `alpha⏎<b/>` / `<b/>⏎beta` (element-adjacent) | stripped |
| `a⏎${x}` (expression-adjacent) | stripped |
| `<b/>   <i/>` (ws-only, one line) | `"   "` preserved |
| `<b/>⏎<i/>` (ws-only, across a newline) | removed |

The scoping is what keeps this safe: only whitespace flanked by text on both sides is preserved, so it always lives inside a single text run and can never inject a whitespace text node between elements (which would cause inline-element gaps or invalid `<table>`/`<tr>` children). Element-, expression-, and edge-adjacent whitespace is stripped as layout, unchanged.

## Tests
Rewrote the `#359` regression and whitespace blocks in `test/jsx-tag.ts` to assert the preserve behavior. Full suite (598) passes; `tsc --noEmit`, ESLint, and Prettier clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@
   proxy-based hyperscript DSL) no longer causes circular references. The JSX
   automatic runtime also only assigns `key` when it is present.
 
+- **Fix `jsx` template dropping the break between adjacent text lines** (#359)
+  Whitespace between two text runs is now preserved verbatim, so multi-line
+  prose no longer loses the break at its line ends (`jsx\`<p>alpha⏎beta</p>\``
+  yields `alpha⏎beta`, not `alphabeta`). Unlike JSX, which collapses the run to a
+  single space, the template keeps the author’s whitespace (newlines, blank
+  lines, and the indentation between text): it renders as a space in normal flow
+  and as a real line break in a `<pre>`. Whitespace adjacent to an element, an
+  expression, or the template edge is still stripped as layout.
+
 ## [0.7.9] - 2026-03-31
 ### Performance
 - **Make User Timing API profiling opt-in via `setProfiling()`**
--- a/src/jsx-tag.ts
+++ b/src/jsx-tag.ts
@@ -183,6 +183,12 @@ export function parse(spans: ArrayLike<string>): ParseResult {
 									//   jsx` \
 									//   `
 									before = before.slice(0, -1);
+								} else if (before && end < span.length && span[end] !== "<") {
+									// Whitespace between two text runs is preserved verbatim, unlike
+									// JSX, which collapses it to a space. Whitespace touching an
+									// element, expression, or template edge is layout and gets
+									// stripped below, hence the guard: text before, text after.
+									before += newline;
 								} else {
 									before = before.replace(/\s*$/, "");
 								}
--- a/test/jsx-tag.ts
+++ b/test/jsx-tag.ts
@@ -62,6 +62,76 @@ test("newlines and whitespace", () => {
 	);
 });
 
+test("text-text newlines are preserved as a newline (#359)", () => {
+	// A newline between two text runs is preserved as a single newline. Unlike
+	// JSX, which collapses it to a space, the template has no formatter reflowing
+	// it, so the line break is the author's intent: it renders as a space in
+	// normal flow and as a real line break in a `<pre>`.
+	Assert.equal(
+		jsx`<p>alpha
+beta</p>`,
+		createElement("p", null, "alpha\n", "beta"),
+	);
+	// Everything between two text runs is preserved verbatim: the blank line
+	// stays two newlines, and the indentation before `three` is kept.
+	Assert.equal(
+		jsx`<p>one
+two
+
+   three</p>`,
+		createElement("p", null, "one\n", "two\n\n   ", "three"),
+	);
+	// A newline adjacent to an element is still stripped as layout: only the
+	// text-text break is kept.
+	Assert.equal(
+		jsx`<p>alpha
+<b>x</b>
+beta</p>`,
+		createElement("p", null, "alpha", createElement("b", null, "x"), "beta"),
+	);
+});
+
+test("preserves significant whitespace verbatim (diverges from JSX collapse)", () => {
+	// Interior spaces within a line are preserved, not collapsed.
+	Assert.equal(jsx`<p>a   b</p>`, createElement("p", null, "a   b"));
+	// A newline before an expression is stripped as layout — the expression
+	// boundary is structural, like an element.
+	Assert.equal(
+		jsx`<p>a
+${"X"}</p>`,
+		createElement("p", null, "a", "X"),
+	);
+	// Leading whitespace on the first line of a text run is preserved (it is not
+	// indentation following a newline); the break to the next line is kept.
+	Assert.equal(
+		jsx`<p>  Hello
+World</p>`,
+		createElement("p", null, "  Hello\n", "World"),
+	);
+	// Whitespace-only text between elements is preserved on a single line ...
+	Assert.equal(
+		jsx`<p><b>x</b>   <i>y</i></p>`,
+		createElement(
+			"p",
+			null,
+			createElement("b", null, "x"),
+			"   ",
+			createElement("i", null, "y"),
+		),
+	);
+	// ... and removed entirely when it spans a newline.
+	Assert.equal(
+		jsx`<p><b>x</b>
+<i>y</i></p>`,
+		createElement(
+			"p",
+			null,
+			createElement("b", null, "x"),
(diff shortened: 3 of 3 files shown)


brainkim:
## Whitespace parity investigation (for the record)

Before merging I wanted to know whether the `jsx` tag is now *exactly* JSX-faithful, so I stopped reasoning from Babel's algorithm and measured against ground truth: every case below was transpiled through **both** reference transpilers the issue cites — esbuild (`--loader=jsx`) and `tsc --jsx react` — and compared to Crank's output on this branch. esbuild and tsc agree with each other throughout.

### Result: parity on every in-content case

| source | Crank (this PR) | esbuild + tsc |
|---|---|---|
| `alpha⏎beta` (text–text) | `alpha beta` | `alpha beta` |
| `alpha⏎<b/>` / `<b/>⏎beta` (element-adjacent) | stripped | stripped |
| `a   b` (interior spaces) | `a   b` | `a   b` (neither collapses) |
| `a⏎{X}` / `{X}⏎b` (expression-adjacent) | no space | no space |
| `  Hello⏎World` (first-line leading) | `  Hello World` | `  Hello World` |
| `a⇥b` (tab in text) | `a⇥b` | `a⇥b` |
| `<b/>···<i/>` (ws-only, one line) | preserved | preserved |
| `<b/>⏎<i/>` (ws-only, across a newline) | removed | removed |

These are locked in by the `matches JSX whitespace handling (verified against esbuild + tsc)` test block.

### Two findings worth recording
- **Tabs in text are not a divergence.** Babel converts tab→space in text nodes, but esbuild and tsc both **preserve** tabs — same as Crank. So there is nothing to change here; the de-facto reference behavior already matches.
- **Expression-adjacent newlines are not a special case.** JSX strips the newline there too (the text node's only non-empty line is the adjacent one), so Crank's behavior matches rather than being a "conservative" limitation.

### The one intentional difference
**Template-edge whitespace.** Crank trims leading/trailing whitespace at the very start and end of the whole template (`jsx\`⏎  <p/>⏎\``), where JSX-in-a-fragment would preserve it. This stays divergent on purpose: a template body has no enclosing element boundary, and authors pad it with newlines/indentation for readability — preserving it would inject spurious whitespace text nodes into every multi-line template. It's covered by the existing `top-level strings` and `newlines and whitespace` tests.

The escaped-newline `\` affordance is inherent to template literals (no JSX analog) and is kept.

**Bottom line:** #359's fix brings the `jsx` tag to full whitespace parity with both reference transpilers for in-content text; the only remaining difference is the deliberate, necessary template-edge trimming.

brainkim:
Pivoted this PR from **collapse-to-a-space** (JSX parity) to **preserve-the-newline**, and updated the body/title/tests to match.

The original approach matched JSX: a text–text newline → single space. But digging into *why* JSX has that rule, it turns out to exist purely to keep source safe under a reflowing formatter — Prettier/tsc rewrap JSX text to a print width, so a literal newline can't be allowed to carry meaning. That rationale doesn't hold for a tagged template: **Prettier has no embedded formatter for the `jsx` tag and leaves the template interior byte-for-byte**, so nothing reflows it. The collapse rule buys nothing here while costing expressiveness — you can't get a literal newline, a blank line, or a multi-line `<pre>` out of literal template whitespace.

So preserve is the better fit: more faithful (a `<pre>` renders correctly), more expressive, and lossless-recoverable for the pixel-exact case via `${...}` — the same move JSX makes with `{"\n"}`. Structural whitespace (indentation, element/expression-adjacent) is still stripped as layout. 598 tests green.

brainkim:
Follow-up (`f5d149d3`): tightened preserve to keep whitespace between text runs **verbatim**, not flattened to a single newline.

The prior commit still collapsed a run to one `\n`, which dropped blank lines and the indentation between text — a half-measure that quietly reintroduced the very collapsing this PR removes. Now the whitespace run is appended as-is.

Why it's safe to go this far: the branch only fires when whitespace is flanked by text on *both* sides, so it always lands inside a single text run. It can never emit a whitespace text node *between elements*, which is the thing that would cause inline-element gaps or invalid `<table>`/`<tr>` children — that hazard is confined to element-adjacent whitespace, which stays stripped. Bonus: a `<pre>`'s internal indentation now survives, so preformatted blocks render correctly without `${...}`.

PR #365, opened 2026-06-24
Add the `before` lifecycle hook (#324)

### LINKED ISSUE #324: Add `before` lifecycle hook
## Summary

Add a `before` lifecycle hook that fires before a component re-renders. This enables early cancellation of in-flight async work that's about to become stale.

## Motivation

When a component re-renders, any in-flight async operations from the previous render become stale. While Crank already discards stale render results (newer renders win), the async work itself continues running until completion - consuming bandwidth, CPU, and other resources unnecessarily.

A `before` hook would allow:
- Early cancellation of fetch requests via AbortController
- Cleanup of expensive computations before they complete
- Integration with AbortSignal-based APIs

## Possible API

```tsx
async function *UserProfile({userId}) {
  let controller = new AbortController();

  for await (const {userId} of this) {
    this.before(() => {
      controller.abort();
      controller = new AbortController();
    });

    yield <div>Loading...</div>;

    const user = await fetchUser(userId, {signal: controller.signal});
    yield <div>{user.name}</div>;
  }
}
```

## Open Questions

- Callback API (`this.before(fn)`) vs other approaches?
- Relationship to existing lifecycle: `schedule`, `after`, `cleanup`
- Could eventually be exposed as event on component's EventTarget? (see #318)

brainkim:
Closes #324.

Adds **`Context.before(callback)`** — a lifecycle hook that fires immediately before a component re-renders, and *only* on re-renders (not the initial render). It lets a component cancel in-flight async work from the previous render before it becomes stale.

## Use case
```tsx
async function *UserProfile(this: Context, {userId}) {
  let controller = new AbortController();
  for await ({userId} of this) {
    this.before(() => {
      controller.abort();
      controller = new AbortController();
    });
    yield <div>Loading…</div>;
    const user = await fetchUser(userId, {signal: controller.signal});
    yield <div>{user.name}</div>;
  }
}
```
When `userId` changes, `before` fires and aborts the stale fetch before the loop re-iterates.

## Semantics
- Fires **once per registration** — re-register on each render, like `schedule`.
- **Never itself triggers a re-render** (unlike `refresh`/`schedule`/`after`).
- Fires **synchronously when a re-render is requested**, not when the component re-runs — so the abort is prompt even while the component is blocked on the very async work being cancelled.

## Implementation
A `beforeMap` keyed by context state (mirroring `scheduleMap`/`cleanupMap`), plus `runBeforeCallbacks(ctx)` invoked at the two synchronous re-render request points: `diffComponent`'s update branch (prop changes / parent re-renders) and `refresh()`. Callbacks are cleared after firing.

## Tests
`test/before.tsx` covers: no-fire on initial render; ordering before re-render (refresh + prop change); once-per-registration; no self-triggered re-render; the AbortController use case; and the async-generator `for await` path. Full suite green (580); `tsc --noEmit` clean.


src/crank.ts
```diff
@@ -2287,8 +2287,24 @@ const provisionMaps = new WeakMap<ContextState, Map<unknown, unknown>>();
...
 
+const beforeMap = new WeakMap<ContextState, Set<Function>>();
+
 const cleanupMap = new WeakMap<ContextState, Set<Function>>();
 
+// Fires the registered before() callbacks for a component that is about to
+// re-render, then clears them (callbacks re-register on each render, like
+// schedule). Invoked synchronously when a re-render is requested — not when the
+// component re-runs — so in-flight async work can be aborted promptly.
+function runBeforeCallbacks(ctx: ContextState): void {
```
brainkim: Probably could be renamed to before(), similar to the other methods, and the commentary is not helpful

src/crank.ts
```diff
@@ -2799,6 +2838,10 @@ function diffComponent<TNode, TScope, TRoot extends TNode | undefined, TResult>(
 				return diffComponent(adapter, root, host, parent, scope, ret);
 			});
 		}
+
+		// A mounted component is about to re-render (e.g. from a prop change or a
```
brainkim, line 2848: Is there a single call-site so we can inline this function?

brainkim (commented):
@claude some thoughts

src/crank.ts
brainkim: Done in 64bb290d — renamed to `before(ctx)` and dropped the comment.

src/crank.ts
brainkim: Two call-sites: the explicit `refresh()` path and the mounted re-render in `diffComponent`. The initial-mount branch immediately below deliberately does *not* fire it, so folding this into `enqueueComponent` would wrongly run before-callbacks on first mount. Kept as a helper, renamed to `before()`.

docs/SKILL.md
```diff
@@ -131,6 +131,7 @@ import {jsx, html, Fragment, renderer, domRenderer, htmlRenderer, DOMRenderer, H
 
 ```js
 this.refresh(callback?)   // Mutate state and re-render
+this.before(callback)     // Run before the next re-render (e.g. abort stale async work)
```
brainkim: @claude One of these methods is not like the other one (zero-arg promise form)

brainkim: Right — verified in the impl. `schedule`/`after`/`cleanup` each carry a second overload where the zero-arg call returns `Promise<TResult>` and resolves *with the rendered value* at that lifecycle moment (their callbacks take `(value: TResult)`). `before` is `before(callback: () => unknown): void` — single signature, and its callback takes no value, because it fires before any render exists. So there is no `TResult` to resolve a promise with; a `before(): Promise` could only resolve with `void`, which is the asymmetry rather than a fix for it. (And `await this.before()` would largely duplicate `for await...of this` as a “wait for the next update” signal.)

So it is a genuine fork, your call:
1. Keep it callback-only and make the doc asymmetry intentional (drop the implied `?`, maybe a one-word note).
2. Add `before(): Promise<void>` for surface symmetry, accepting the void resolution.

The SKILL.md line is accurate to (1) as written. Say which way and I’ll align the doc (and the overload if you want 2).

brainkim: Done in abb5f9fd — added the `before(): Promise<void>` overload (mirrors the `schedule` idiom: `new Promise((resolve) => this.before(() => resolve()))`). Updated the SKILL.md signature to `before(callback?)`, the CHANGELOG, and added a test asserting it stays pending until a re-render then resolves with `undefined`. The lone remaining asymmetry — resolving with `void` rather than the rendered value — is inherent, since nothing has rendered yet.

brainkim:
Closing this in favor of #374 — the design moved a lot in review and I'd rather land a clean PR for it.

Two things drove the change. First, `before` was always bugging me: before *what*? It names the timing, not what the hook is for. Second, once I looked at the actual use case — aborting stale async work — the hook is really about *staleness*, and a render goes stale whether it's superseded by a re-render **or** removed on unmount. Firing only on re-render meant you'd need `before` + `cleanup` to cover both cases, which is silly for one intent.

So #374 is `stale`: fires on retirement (re-render or unmount) from a single callback, and hands you a promise for the successor's result (or `undefined` on unmount), so you get both the pre-emptive abort and the actually-replaced moment. Same #324 feature, better shape. Superseded by #374.

Issue #293, opened 2025-07-27
Implement streaming HTML renderer API with SSR support


brainkim:
## Streaming HTML Renderer for SSR
Allow streaming server-side rendering (SSR) in Crank.

### Details
Add an API for streaming HTML rendering, likely via a method such as `HTMLRenderer.render` that accepts a `WriteableStream` as the second argument.

```js
const result = await renderer.render(<Page />, stream);
```

The result would still be a string, but the stream would be written eagerly as elements fulfilled.

### Motivation
Streaming SSR is important for performance and scalability, especially for large or dynamic applications. Most HTML pages will have a static head, and async body, so it’s imperative we send that head early for a good TTFB and to allow clients to fetch assets eagerly.

### API Example
```js
import { renderer } from "@b9g/crank/html";

self.addEventListener("fetch", (event) => {
  const { readable, writable } = new TransformStream();

  // Write to the stream
  renderer.render(<App />, writable);

  // Respond with the readable side
  event.respondWith(new Response(readable));
});
```

### Implementation details
I used to think that this could be implemented using the internal renderer methods, namely scope, which is called in a pre-order traversal of the tree, but I’m realizing that none of the renderer methods are called in a depth-first order because of async siblings. The streaming SSR algorithm can be reduced to the following problem:

```ts
// Assignment: Eager DFS Traversal of an Async Tree (Enter/Exit)

/**
 * A tree-like structure where each node is one of:
 * - A string: a leaf value.
 * - An array: representing ordered child nodes.
 * - A function: when called, returns a Promise that resolves to a subtree.
 */
type Tree = string | Tree[] | (() => Promise<Tree>);

/**
 * Walk a tree, recursively executing all functions found in the tree as soon
 * as possible.
 *
 * @param tree - The tree to walk.
 * @param onEnter - Called when visiting a string or array node (pre-order).
 * @param onExit - Called after fully traversing a string or array node (post-order).
 *
 * @returns Concatenation of all strings in DFS order:
 * - Returns `string` if the tree contains no async nodes.
 * - Returns `Promise<string>` otherwise.
 */
function walk(
  tree: Tree,
  onEnter: (value: string) => void,
  onExit: (value: string) => void
): Promise<string> | string {
  throw new Error("Not implemented");
}

/*
Requirements
------------
1. Delay nodes (functions) must execute as soon as encountered:
   - All sibling delays are started immediately in array order.
   - If a delay resolves to another delay, that nested delay is also started
[...]

PR #367, opened 2026-06-25
Streaming HTML rendering via a polymorphic render() (#293)

brainkim:
## Streaming HTML rendering through the real renderer (#293)

Makes `htmlRenderer.render()` stream. The static shell flushes before async content settles, and async parts emit in document order as they resolve — implemented in the **actual core renderer**, not a parallel path.

```js
htmlRenderer.render(<App />);                  // string | Promise<string>  (unchanged)
htmlRenderer.render(<App />, writableStream);  // Promise<string>, streamed to the sink
```

`render()` always resolves to the full HTML string — its return type is `string | Promise<string>` regardless of the second argument. A writable sink only adds incremental flushing as a side effect; callers who want a streaming `Response` wrap the sink themselves.

### The idea

Crank already splits rendering into **diff** (fan-out + `Promise.all` join) and **commit** (`adapter.arrange`, bottom-up). The atomic path resolves the whole subtree before committing — `arrange` is monolithic (`open + children + close` at once), which is exactly why the shell can't flush early.

Streaming is the same diff with a different *join/commit*: instead of awaiting the whole `Promise.all`, walk the resolving retainer tree in document order, emitting each completed prefix. The unification is two emission points on one traversal:

- **exit** (subtree complete) → `arrange`. The DOM and atomic HTML read this; rollback-safe, unchanged.
- **enter/exit** → two new optional adapter hooks, `open` and `close` (the enter/exit split of `arrange`): the opening node is emitted before the children, the closing node after. Each mints a single `TNode` like `create`/`text`/`raw`, so `read` stays the only projection to the result type.

So **atomic is the buffered special case of streaming** — same machinery, the DOM just listens on exit only.

### What changed

- `Renderer.render()` is now polymorphic: a writable sink (duck-typed `getWriter`, gated on `open`/`close`) routes to the streaming commit and resolves to the full result; otherwise the untouched atomic path.
- New core: `renderRootStream` / `commitStream` — incremental, in-order commit that awaits each child as it reaches it. Hosts emit their open tag synchronously and recurse; only unresolved async components are awaited (via their inflight *block*, so a component's own shell streams while its descendants are still pending).
- New optional adapter hooks `open`/`close` (mint the opening/closing nodes that bracket the children); `printOpen`/`printClose` extracted from html's `arrange` (byte-identical) as their primitives. `arrange` itself is unchanged — decomposing it is a breaking change for a future major, not this PR.

### Tested (all green; full suite unchanged for the non-streaming path)

`renderer.render(el, sink)` asserted **byte-identical to `renderer.render(el)`** across host/attrs/void/innerHTML/fragment/svg/function/generator/nested/async; shell-flush; sync-component shell-flush; **async-component** shell-flush before descendants; document-order async siblings; throwing-async-component rejects; sync error boundary recovers.

### Known limitations (documented in tests)

- **Component lifecycle callbacks don't fire while streaming.** `this.schedule`/`this.after`/`this.cleanup` fire in atomic SSR but not in the streaming path (`commitStream` recurses component children directly and never calls `commitComponent`). A silent divergence from `render()` — needs a decision: fire them, or document the difference.
- **No backpressure.** Writes are fire-and-forget (`writer.write` isn't awaited), so a slow sink queues chunks in memory. Fine for many setups; a follow-up for production-grade streaming.
- **Streams for latency, not memory.** Because `render()` resolves to the full result, the whole output is still buffered to the end — this wins time-to-first-byte, not bounded server memory. A future void-returning variant could drop the accumulator.
- **Async error past a boundary isn't recoverable mid-stream** — by the time it rejects, earlier siblings have flushed, so the fallback can't replace it; the stream rejects. (React solves this with abort-and-swap; out of scope here.)
- The DOM renderer leaves `open`/`close` undefined and is entirely unaffected.

### Scope

Streaming shares the **diff/execution** phase with the atomic renderer (components run once through the real machinery — hence byte-identical output and matching async/Suspense semantics) but uses a **separate commit** (`commitStream`) instead of `commit`. Every limitation above traces to that one fork: `commitStream` doesn't go through `commitComponent`, so lifecycle/refs don't fire, and the two commit paths must be kept in sync by hand.

Fully unifying this — making the *real* `commit` incremental so there's a single traversal parameterized by flush-as-you-go vs. buffer-to-end, with lifecycle firing identically in both — is the actual end state. It's a deep hot-path refactor (touches scheduling/hydration/fallbacks) and is deliberately **out of scope** here. This PR ships the forked, experimental version; unification is follow-up.

### Notes for review

The first two commits build a standalone prototype; it's deleted once the core path subsumes it. History is kept to tell the story — squash on merge if preferred. This is 0.9-scoped streaming, behind an additive, opt-in surface.


brainkim:
Closing in favor of an integrated rework on `feat/streaming-integrated` (new PR to follow).

This PR's parallel construction — a separate `commitStream` path plus `open`/`close` adapter hooks — bypasses the renderer's own `scope`/`create`/`patch`/`arrange` methods, which forced lifecycle, `flush`, and `unmount` to be re-added by hand. #293 asks for streaming "implemented using the internal renderer methods." The replacement does exactly that: streaming rides the renderer's existing commit traversal as a single maybe-async walk, with atomic rendering as the degenerate (fully-resolved, no-sink) case.

Issue #369, opened 2026-07-09
Decide on modernizing the UMD build and default stack


brainkim:
Tracking issue for a **0.8 decision** — parking it here so it isn't lost.

The UMD build and the default module/entry stack have accrued cruft and are due for a modernization pass. During the 0.8 cycle, decide:

- Whether to keep, slim, or drop the **UMD build** (vs. ESM-first / modern bundleless distribution).
- What the **default stack / entry points** should be going forward.
- Back-compat and migration implications for existing consumers.

Raised during 0.8 planning; specifics TBD.

PR #370, opened 2026-07-09
Migrate build to libuild; rework the UMD global surface

brainkim:
Migrates the build to [`@b9g/libuild`](https://github.com/bikeshaving/libuild) (matching the other @b9g libraries) and reworks the UMD/CDN global surface. Part of #369.

## Build tooling
- Deletes `rollup.config.js` and the `dts()` / `copyPackage()` / `rewritePaths()` hacks. libuild derives entry points, the `exports` map, `.d.ts`, and the clean dist `package.json` from convention.
- Drops `rollup`, `rollup-plugin-typescript2`, `magic-string`, `shx`; adds `@b9g/libuild`. Scripts become `build: libuild build` / `publish: libuild publish` with the standard `prepublishOnly` guard.
- `tsconfig.json`: enable `declaration`, drop `noEmit`, so libuild emits types (the `typecheck` script still passes `--noEmit` explicitly).
- Output moves `dist/*` → `dist/src/*`, but **public specifiers are unchanged** (`@b9g/crank`, `/dom`, `/html`, …) because libuild regenerates the exports map. No consumer-facing path break.
- Adds `src/jsx-dev-runtime.ts` so `./jsx-dev-runtime` survives as a real entry (it was a hand-written alias before; libuild only maps real files).

## UMD global (`window.Crank`) — **breaking**
`src/umd.ts` now mirrors `standalone` but **pins `renderer` to the DOM renderer** (a `<script>` global only runs in the DOM), and exposes the tagged templates for no-build authoring:

| global | is |
|---|---|
| `createElement`, `Fragment`, `Portal`, `Copy`, `Text`, `Raw`, `Element`, `isElement`, `cloneElement`, `Context`, `Renderer`, `setProfiling` | core (flat) |
| `Crank.jsx`, `Crank.html` | tagged templates |
| `Crank.renderer`, `Crank.domRenderer` | DOM renderer |
| `Crank.DOMRenderer` | DOM renderer class |
| `Crank.htmlRenderer`, `Crank.HTMLRenderer` | server HTML renderer / class |

**Breaking:** `Crank.html` was the HTML-string renderer; it is now the `html` tagged template. The `Crank.dom.*` / `Crank.html.*` namespaces are gone in favour of the flat names above.

## Verified locally
- `libuild build` → ESM + CJS + UMD (global auto-named `Crank`), `.d.ts` for every entry with `declare global { namespace Crank / JSX }` augmentation preserved, regenerated `exports` map (incl. `./jsx-dev-runtime`).
- `tsc --noEmit` clean; **596 tests pass**; ESLint clean. `dist/` stays gitignored.

## Open items (the "hard boss" surfacing libuild gaps)
1. **CDN fields dropped.** libuild's generated dist `package.json` strips `unpkg`/`jsdelivr`, so `unpkg.com/@b9g/crank` still resolves to `main` (CJS), not the UMD global. Either libuild needs to preserve/emit those fields, or we document the explicit `.../dist/src/umd.js` path. Left out of this PR since hand-adding them to the source manifest is inert.
2. **`.d.ts` needs a real local install** (not `bunx`) for libuild to resolve `tsc` — expected, just noting it for CI.


package.json
```diff
@@ -17,101 +20,12 @@
...
     "playwright-test": "^14.1.9",
     "prettier": "^3.5.3",
-    "rollup": "^4.39.0",
-    "rollup-plugin-typescript2": "^0.36.0",
-    "shx": "^0.3.4",
     "sinon": "^17.0.2",
     "typescript": "^5.8.3",
     "uvu": "^0.5.6"
   },
-  "exports": {
```
brainkim, line 30: Where did the exports map go?

brainkim (commented):
@claude I wonder if we can make this seamless and have the files be exactly the same, but other than that this is pretty exciting.

package.json
brainkim: libuild generates the exports map (plus `main`/`module`/`types`) from `src/*.ts` at publish time and writes it into the *published* package.json — a hand-written one in source would just drift. You can see the generated result with `npm pack` / a publish dry-run.

brainkim: Correcting my earlier answer: crankdown *does* commit its generated exports map, so for consistency #370 should too. Regenerated with `libuild build --save` in ee894779 — the full map is now in source (`.`, bare, and `.js` variants per entry, plus `./package.json`), matching crankdown.

The build script stays `libuild build` (not `--save`), same as crankdown: the map is materialized once and committed, and libuild warns on stale exports if a `src/` entry is added later without re-saving. The published `dist/package.json` is identical either way — `--save` only affects whether the map is *also* visible in the repo.

(Separately, re your flat-`dist/` question: not possible without a libuild change — `distSrcDir` is hardcoded to `dist/src` with esbuild `outbase: cwd`, so outputs mirror the `src/` path. Would need an exposed `outbase`/flat option upstream.)

PR #375, opened 2026-08-12
Replace playwright-test/uvu with the libuild test runner

brainkim:
Moves the core suite from `playwright-test --runner uvu` to `libuild test`, so crank consumes libuild's test stack alongside its build stack.

Verified on a clean install of libuild 0.2.17, no patching: **602 passed, 3 skipped, exit 0.**

## What changed

| before | after |
|---|---|
| `uvu` `suite()` | `describe()` |
| `test.before.each` / `test.after.each` | `beforeEach` / `afterEach` |
| `uvu/assert` (`Assert.is`, …) | `expect(...)` |
| `playwright-test --runner uvu` | `libuild test test/ --platform <browser>` |
| `playwright-test` (bundled browsers) | `playwright` (explicit `playwright install`) |

`playwright-test.config.js` is deleted; CI's browser matrix (chromium/firefox/webkit) is unchanged and now drives `--platform`.

Each file is wrapped in `describe()` rather than left flat. That is not cosmetic: uvu's `suite()` scoped hooks per file, but libuild bundles all 30 files into one root suite, so top-level `beforeEach` hooks would run before *every* test in the suite. Left flat, all 602 tests failed with Sinon's "Attempted to wrap error which is already wrapped" — 26 files' hooks each stubbing `console.error` on top of one another.

## Three bugs the conversion shook out

**`test/jsx-tag.ts` has never run.** `playwright-test.config.js` declared `input: ["test/*.tsx"]`, which does not match the one `.ts` test file. Its 25 tests were silently excluded from every CI run. Including them takes the suite from 577 to 602. They pass.

**One assertion was never actually checked.** uvu's `throws(fn, expects, msg)` treats a *string* second argument as the failure message, not a matcher — only a RegExp or function is compared. So:

```js
Assert.throws(() => renderer.render(<Component />, document.body), "Context iterated twice");
```

asserted only that *something* threw. `expect().toThrow(string)` does substring-match, which surfaced that the real message is `<Component> context iterated twice without a yield` — lowercase, and differently worded. The expected string is corrected.

**Two spots leaned on assertion signatures.** uvu types `ok(val): asserts val`, so `Assert.ok(x instanceof Element)` narrowed `x` for later lines. `expect().toBeTruthy()` does not narrow, so `mathml.tsx` and `svg.tsx` now access those values explicitly.

## Fidelity

- 602 `test()` declarations and 3 `test.skip`s before and after — nothing added or lost. The run reports exactly `602 passed, 3 skipped`.
- Assertion messages, which `expect()` has no parameter for, are kept as block comments instead of being dropped (`expect(x).toBe(1) /* cleanup should be called */;`).

## libuild bugs this found, fixed in 0.2.17

This PR was blocked on two of them; both are now fixed upstream and the workarounds are gone.

1. **The browser runner started before any test registered.** `@b9g/libuild/test` picks its backend with top-level await, making it an async module, so every importing test file's body is deferred past the runner's `queueMicrotask` start. It ran with zero tests and reported a green empty run. Fixed by having the generated entry set a ready flag as its last statement, which ESM guarantees runs after every imported file has evaluated.

2. **A zero-test run exited 0.** Merging this against 0.2.16 would have silently disabled crank's entire suite with CI green. Now a hard failure — verified here: an empty selection exits 1 with "test files were bundled and loaded, but zero tests registered".

Also fixed in 0.2.17 and used here: `test.skip` exists in the browser runner (the three disabled hydration tests use it again instead of a local no-op, and are reported as skipped), and `libuild test test/` discovers files correctly when the root is itself named `test/` (so discovery is the plain directory form rather than explicit globs).

## Verification

`602 passed, 3 skipped, 0 failed` on chromium against a clean install with no patching, exit 0. `tsc --noEmit` clean, eslint clean. A deliberately empty selection exits 1.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Install browser
         run: bunx playwright install --with-deps ${{ matrix.browser }}
       - name: Run core tests
-        run: bunx libuild test "test/*.tsx" "test/*.ts" --platform ${{ matrix.browser }}
+        run: bunx libuild test test/ --platform ${{ matrix.browser }}
       - name: Run eslint plugin tests
         run: bun run test:eslint
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,11 +43,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: oven-sh/setup-bun@v2
       - run: bun install
-      - name: Install WebKit dependencies
-        if: matrix.browser == 'webkit'
-        run: bunx playwright install-deps webkit
+      - name: Install browser
+        run: bunx playwright install --with-deps ${{ matrix.browser }}
       - name: Run core tests
-        run: bun run test:core -- --browser ${{ matrix.browser }}
+        run: bunx libuild test "test/*.tsx" "test/*.ts" --platform ${{ matrix.browser }}
       - name: Run eslint plugin tests
         run: bun run test:eslint
 
--- a/bun.lock
+++ b/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "@b9g/crank",
       "devDependencies": {
-        "@b9g/libuild": "0.2.18",
+        "@b9g/libuild": "^0.2.19",
         "@types/node": "^22.14.0",
         "@types/sinon": "^17.0.4",
         "@typescript-eslint/eslint-plugin": "^8.29.1",
@@ -21,7 +21,7 @@
         "typescript": "^5.8.3",
       },
     },
-    "packages/astro-crank": {
+    "node_modules/.bun/@b9g+crank@root/node_modules/@b9g/crank/packages/astro-crank": {
       "name": "astro-crank",
       "version": "0.3.0",
       "peerDependencies": {
@@ -29,7 +29,7 @@
         "astro": "^4.0.0 || ^5.0.0",
       },
     },
-    "packages/crank-codemods": {
+    "node_modules/.bun/@b9g+crank@root/node_modules/@b9g/crank/packages/crank-codemods": {
       "name": "@b9g/crank-codemods",
       "version": "0.1.0",
       "dependencies": {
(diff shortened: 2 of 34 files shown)


brainkim:
`@b9g/libuild@0.2.17` is published and should unblock this draft — both blockers are fixed, along with the minor note:

**Blocker 1 (runner fires before registration):** fixed, but not with the `setTimeout` patch — any scheduler-based start still loses to a test file whose top-level await crosses a macrotask (its later tests silently vanish from a green run). Instead the generated entry now sets a ready flag as its *last statement* — ESM guarantees that runs only after every imported file, TLA continuations included, has fully evaluated — and the runner waits on it. Your 602-passing conversion was produced under the equivalent patch, so it should pass unchanged; the local libuild patch can be dropped.

**Blocker 2 (zero tests exits 0):** a browser run that discovers files but registers zero tests now exits 1.

**Minor (`libuild test test/` finds nothing):** fixed — when the discovery root itself is named `test`/`tests`/`__tests__`, files without the `.test.` infix (e.g. `test/dom.tsx`) are discovered. The explicit-glob invocation in this PR also still works.

Two things that may affect the diff here:

- The browser runner now has `test.skip`/`test.todo`/`describe.skip` (they register nothing and tally as skipped), so the local no-op `skip()` shim for the three hydration tests can go.
- Browser runs are stricter than what this PR was tested against: **uncaught page errors now fail the run** (previously they were log lines while the run stayed green), and a bundle that breaks before the runner starts is a reported failure. If anything in the suite was quietly relying on a swallowed page error, it will surface as red — that's the fix working, not a regression.

Full details in the release notes: https://github.com/bikeshaving/libuild/releases/tag/v0.2.17

brainkim:
The webkit CI failure here is a real, pre-existing crank bug that this PR's runner made visible (it installs current webkit; `playwright-test` pinned webkit-2191, which predates the JSC change that exposes it). Fix is #379, with the full engine forensics in [this comment](https://github.com/bikeshaving/crank/pull/379#issuecomment-5275501595) — short version: WebKit's C++ Promise rewrite ([JetStream 3 post](https://webkit.org/blog/17899/introducing-the-jetstream-3-benchmark-suite/)) made `.finally()` two ticks faster, and crank's enqueue machinery raced on those ticks.

Once #379 lands on main, a rebase here should turn webkit green. The 22 extra failures in the webkit job are amplification from bikeshaving/libuild#25 (afterEach skipped on test failure → leaked Sinon stubs); firefox is addressed upstream for libuild 0.2.19.

PR #379, opened 2026-08-12
Fix stale children committing when refresh() supersedes an async generator yield

brainkim:
Fixes the webkit failure discovered on #375.

## The bug

When an async generator component calls `refresh()` before its next `yield` has committed, the yielded children are stale: the refresh immediately resumes the generator, and only the fresh children should commit.

```js
async function* Component(this: Context) {
  await Promise.resolve();
  this.refresh();
  yield <span>Hello</span>;    // stale before it commits
  yield <span>Goodbye</span>;  // what should commit
}
```

Whether that held depended on a race between two unsynchronized microtask chains: the parent's commit walk vs. the enqueued run's diff. On V8 and older JavaScriptCore the enqueued diff won (Hello never appears — the behavior the cascades test asserts). On JSC as of Safari 26 the commit walk wins: **Hello flashes into the DOM and `render()` resolves with it.**

## Root cause, measured

Pure-JS probe, no crank involved:

| shape | V8 | old JSC (webkit-2191) | new JSC (webkit-2336) |
|---|---|---|---|
| `Promise.resolve().finally().then()` | tick 4 | tick 4 | **tick 2** |
| `catch().finally().then()` | tick 5 | tick 5 | **tick 3** |

JSC optimized `.finally()` by two microtask ticks. Crank's enqueue machinery (`block.finally(advance)`, `inflight[0].finally(runComponent)`) sits exactly on those ticks, so the race flipped with no crank code changing. Generator resumption and `await` hops are identical across all three engines — it is `.finally` alone.

Main's CI cannot see this: `playwright-test` pins playwright-core 1.54 → webkit-2191. #375's CI installs current webkit → 2336, which is how it surfaced. It reproduces deterministically (3/3 locally), file run alone.

## The fix

Make the supersede explicit instead of racing for it:

- `refresh()` sets a new `IsSupersededByRefresh` flag when a run is already in flight — self-refresh origin only.
- When the pending iteration arrives with the flag set (and not `done`), the stale children are not diffed at all, and the parent's diff resolves with the enqueued run's diff.
- The block chain derives from the iteration via a closure rather than from the parent diff — the enqueued run only starts once the block settles, so deriving it from the now-chasing diff would deadlock.

Two constraints shaped this, both enforced by existing tests:

- **External updates must not chase.** A re-render requested while a run is in flight coalesces into the enqueued run, but the in-flight run commits its own children (`"for...of enqueues"`: p1 resolves showing "Hello 1" while p2–p5 coalesce). The flag keys on refresh origin for exactly this reason.
- **The promise graph must be unchanged when no supersede occurs.** Earlier drafts added one microtask hop to every async-gen diff and broke tests whose assertions sample the DOM mid-flight. The final shape adds zero reactions on the clean path.

## Verification

- chromium (playwright-test, main's runner): **577 passed**
- webkit-2336, webkit-2191, chromium (libuild runner on the #375 branch with this `crank.ts`): **602 passed, 0 failed** each
- Instrumented MutationObserver trace: identical on V8 and new JSC — a single commit of the fresh children; the stale children never appear
- `tsc --noEmit` clean, eslint clean

The 22 other webkit failures on #375 were fallout from this one (libuild's browser runner runs `afterEach` inside its `try`, so the failing test leaked a Sinon `console.error` stub into every later test that stubs — being reported to libuild separately).

Intended for a 0.7.11 patch release via `release/0.7` as well; this is the main-line landing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,18 @@
   renderer works with custom DOM implementations (e.g. termdom) which don’t
   install the browser globals.
 
+- **A self-refresh() during an async generator's execution no longer flashes the superseded children into the DOM on newer Safari.**
+  When an async generator component calls `refresh()` before its next `yield`
+  has committed, the yielded children are stale: the refresh immediately
+  resumes the generator, and only the fresh children should commit. Crank
+  relied on unspecified microtask ordering for this, and JavaScriptCore as of
+  Safari 26 resolves `.finally()` two ticks faster than V8, which flipped the
+  race: the stale children briefly committed and the render promise resolved
+  with them. The superseded children are now skipped explicitly, on every
+  engine. External re-renders are unaffected: an update requested while a
+  render is in flight still coalesces into the enqueued render, and the
+  in-flight render still commits its own children.
+
 - **`createElement` no longer mutates the caller's props object** (#356)
   Children are spread into a fresh props object instead of being assigned onto
   the passed-in props, so reusing a props object across calls (e.g. a
--- a/src/crank.ts
+++ b/src/crank.ts
@@ -326,6 +326,13 @@ const NeedsToYield = 1 << 15;
 const PropsAvailable = 1 << 16;
 const IsSchedulingRefresh = 1 << 17;
 
+/**
+ * A flag which indicates that refresh() was called while a run of the
+ * component was already in flight, meaning the in-flight run's yielded
+ * children are stale before they have committed.
+ */
+const IsSupersededByRefresh = 1 << 18;
+
 function getFlag(ret: Retainer<unknown>, flag: number): boolean {
 	return !!(ret.f & flag);
 }
@@ -2570,6 +2577,12 @@ export class Context<
 		const schedulePromises: Array<PromiseLike<unknown>> = [];
 		try {
 			setFlag(ctx.ret, IsRefreshing);
+			if (ctx.inflight) {
+				// This refresh supersedes the in-flight run: the component itself has
+				// declared that whatever that run yields is already stale.
+				setFlag(ctx.ret, IsSupersededByRefresh);
+			}
+
 			diff = enqueueComponent(ctx);
 			if (isPromiseLike(diff)) {
 				return diff
@@ -2921,6 +2934,9 @@ function runComponent<TNode, TResult>(
 	const ret = ctx.ret;
 	const initial = !ctx.iterator;
 	const tagName = getTagName(ret.el.tag);
+	// Each run starts unsuperseded; the flag describes the run in flight, and a
+	// rejected diff would otherwise leave it set for the next run.
+	setFlag(ctx.ret, IsSupersededByRefresh, false);
 	if (initial) {
 		setFlag(ctx.ret, IsExecuting);
 		clearEventListeners(ctx.ctx);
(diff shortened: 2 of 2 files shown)


brainkim:
## Color commentary: sourcing the engine change

The PR body says JSC resolves `.finally()` two ticks faster than V8. Here is *why*, traced to the shipped WebKit change.

### What WebKit did

WebKit ran a campaign rewriting its microtask queue into JavaScriptCore and migrating the Promise implementation from self-hosted JS builtins to C++ — described in [Introducing the JetStream 3 Benchmark Suite](https://webkit.org/blog/17899/introducing-the-jetstream-3-benchmark-suite/), which explicitly names **`Promise.prototype.finally`** among the methods moved to native code (`Promise.all`, `race`, `allSettled`, `any`, `finally`, `resolve`/`reject`). The commits are visible landing through late 2025, e.g. [[JSC] Move Promise.resolve and Promise.reject to C++](https://www.mail-archive.com/webkit-changes@lists.webkit.org/msg237084.html). Playwright's webkit-2191 (what `playwright-test` pins, mid-2025) predates the campaign; webkit-2336 (WebKit 26.5) has it — which is exactly the version boundary the bisection in this PR found.

### Why the rewrite changes tick counts

Spec-literal `finally` is expensive *by construction*. [ThenFinally](https://tc39.es/ecma262/#sec-thenfinallyfunctions) calls your callback, then builds `PromiseResolve(C, result)` and chains `.then(valueThunk)` onto it purely to pass the original value through — two intermediate promise constructs, each costing a microtask tick, even when the callback returns nothing. That is why `.finally` costs 4 ticks where `.then` costs 2. JSC's C++ implementation short-circuits that dance for non-thenable callback results, and `.finally` collapses to plain-`.then` cost.

### The change is `finally`-specific, not general adoption

My first hypothesis was that JSC had shipped something like TC39's Stage-1 [faster-promise-adoption](https://github.com/tc39/proposal-faster-promise-adoption) proposal. The probes refute that — promise state adoption is untouched:

| operation | V8 | old JSC (2191) | new JSC (2336) |
|---|---|---|---|
| plain `.then` chain (control) | 2 | 2 | 2 |
| executor adoption `res(Promise.resolve())` | 3 | — | 3 |
| handler adoption `.then(() => Promise.resolve())` | 4 | — | **4 (unchanged)** |
| `.finally().then()` | 4 | 4 | **2** |
| `catch().finally().then()` (crank's block-chain shape) | 5 | 5 | **3** |
| async generator resumption | 3 | 3 | 3 |
| `await` in generator body | identical | identical | identical |

Nothing about generators or `await` moved, which is why the breakage was so surgical: only code sitting specifically on `.finally` tick arithmetic — crank's enqueue machinery (`block.finally(advance)`, `inflight[0].finally(runComponent)`) — reordered.

### The conformance question

Strictly, this is observable divergence from the spec: ThenFinally's intermediate promises are normative steps, and tick counts are observable — this suite is the demonstration. The blog frames the work as pure performance. One could file a WebKit conformance bug, but the history says don't bother building on it either way: V8 changed `await`'s tick count in 2018 and TC39 ratified it retroactively ([ecma262#1250](https://github.com/tc39/ecma262/pull/1250)), and the faster-promise-adoption proposal shows the committee's direction is to bless further reductions. Code that depends on cross-chain tick arithmetic loses eventually, whichever engine moves next — which is why this PR removes the dependency instead of asserting the old counts.

### Related

- Surfaced by #375's CI, which installs current webkit rather than playwright-test's pinned 2191 — main's CI cannot see this bug before that migration lands.
- The 1-real-failure → 23-reported-failures amplification on webkit was a separate runner bug, filed as bikeshaving/libuild#25.
- Backported to the 0.7 line on `release/0.7` for 0.7.11 alongside #356 and #359.

PR #380, opened 2026-08-13
README: rewrite the Motivation prose in the project's own voice

brainkim:
Replaces the "Why Developers Choose Crank" listicle and "The 'Just JavaScript' Promise, Delivered" section with a short "Why Crank?" in the project's register.

That prose arrived in 4d756bc2 ("Why Crank?"), which described itself as "a relatively low effort change via Claude", and it reads like it: marketing-bullet cadence ("Fast", "Future-proof", "Promise, Delivered") and claims pitched at a landing page rather than a README. The replacement says the same true things the way the rest of the project says them — concrete mechanism, no superlatives: state is a variable in generator scope, updates are re-execution, control flow is the lifecycle.

The code example above the section is untouched; it was already right.

Also cherry-picked to `release/0.7` so the 0.7.11 npm tarball ships the same README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,19 @@ async function UserProfile({userId}) {
 await renderer.render(<UserProfile />, document.body);
 ```
 
-### Why Developers Choose Crank
+### Why Crank?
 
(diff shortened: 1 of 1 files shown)


brainkim:
Prose still sucks but it’s a start. When are we actually getting AI super-persuasion? @claude

PR #384, opened 2026-08-18
Cache static subtrees in the jsx template tag

### LINKED ISSUE #383: JSX template tag copy optimization
It would be nice if we added automatic `copy` props to any element or subtree which is static for the JSX template tag. I think we already have all the information we need during the parse, and the `copy="x y z"` syntax is available. Would be nice to see if there’s actually any perf improvements. Although, it might change the semantics of what a render is.

brainkim:
Fixes #383.

## What this does

During `parse()`, a `markStatic` walk marks every element whose subtree contains no expression targets. `build()` then constructs each static subtree once and returns the same element instance on every subsequent call of the template. Since crank already skips re-rendering when `ret.el === child` (the documented element-reuse behavior, see `test/reuse.tsx`), static subtrees are skipped by identity on re-renders — the whole diff, patch, and arrange for those parts just doesn’t run. No `copy` props are involved.

The analysis runs once per template (it’s cached with the `ParseResult`); the per-call cost is a couple of property checks in `build`.

## Why identity instead of auto `copy` props

The issue suggests emitting `copy="x y z"` for static props. That turns out to be unsound: `copy` copies the named props from `oldProps`, but unkeyed reconciliation matches retainers by tag, so the retainer at a position can come from a *different* template with different values for those props:

```js
cond ? jsx`<div class="a"/>` : jsx`<div class="b"/>`
```

If the second template carried an auto `copy="class"`, toggling `cond` would keep `class="a"`. Identity caching doesn’t have this problem — a different template’s element fails the `ret.el === child` check and diffs normally (there’s a test for exactly this).

## Perf

Chromium, 20,000 re-renders of a header/nav/footer-style template with one dynamic text expression (~20 elements):

|  | µs/render |
|---|---|
| main | ~19.2 |
| this PR | ~5.1 |

An all-dynamic template (every element has a dynamic prop or child) measures identically before and after (~6.1µs), so there’s no overhead when nothing is static.

## Semantics

This does change what a render is for templates, as the issue anticipated: static parts of a template are no longer re-patched on every render, so out-of-band DOM mutations to them are no longer overwritten by a re-render. This is the same behavior users already get when they hoist an element out of a component by hand; the template tag now effectively does that hoisting for them. There’s a test encoding this so the change is explicit. The changelog entry notes it under the Unreleased section.

One nice edge: comment expressions (`<!-- ${x} -->`) don’t defeat caching, since comments never reach the element tree.

### DIFF
--- a/src/jsx-tag.ts
+++ b/src/jsx-tag.ts
@@ -61,12 +61,13 @@ export interface ParseElement {
 	// ParseValue is used to represent spread props.
 	props: Array<ParseProp | ParseValue>;
 	children: Array<ParseElement | ParseValue>;
-	// Set in parse() when the subtree contains no expressions. build() returns
-	// the same element instance for static subtrees, which lets renderers skip
-	// them by identity on re-renders.
-	static?: boolean;
-	built?: Element;
 }
+
+// Subtrees which contain no expressions, as marked by parse(). build() creates
+// each static subtree once and returns the same element instance on every
+// call, which lets renderers skip them by identity on re-renders.
+const staticElements = new WeakSet<ParseElement>();
+const builtElements = new WeakMap<ParseElement, Element>();
 
 export interface ParseValue {
 	type: "value";
@@ -649,13 +650,17 @@ function markStatic(el: ParseElement, targets: Set<object>): boolean {
 		}
 	}
 
-	el.static = isStatic;
+	if (isStatic) {
+		staticElements.add(el);
+	}
+
 	return isStatic;
 }
 
 function build(parsed: ParseElement, spans?: ArrayLike<string>): Element {
-	if (parsed.static && parsed.built) {
-		return parsed.built;
+	const built = builtElements.get(parsed);
+	if (built) {
+		return built;
 	}
 
 	if (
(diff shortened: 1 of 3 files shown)


src/jsx-tag.ts
```diff
@@ -61,6 +61,11 @@ export interface ParseElement {
 	// ParseValue is used to represent spread props.
 	props: Array<ParseProp | ParseValue>;
 	children: Array<ParseElement | ParseValue>;
+	// Set in parse() when the subtree contains no expressions. build() returns
+	// the same element instance for static subtrees, which lets renderers skip
+	// them by identity on re-renders.
+	static?: boolean;
```
brainkim: I hate this. The PR says we’re using identity for subtrees. That’s essentially the same as `copy={true}`? Also if we own the JSX parser and cache per template tag call, can’t we cache the elements which are static subtrees?

brainkim (commented):
@claude When I mentioned the copy string prop, I was mainly thinking of the `copy="!children"` case. This would allow a parent to skip re-rendering the props while allowing dynamic children to be updated.

Also very hesitant to add `static` to the element definition. Can we just have the parser cache elements if we want to go with the element reuse path?

brainkim:
Okay, so @claude has convinced me that auto-emitting the `copy` prop is unsound because there isn’t any way to distinguish static subtrees in a JSX template tag call which coincidentally matches the structure of another call:

```js
cond ? jsx`<div class="a" />` : jsx`<div class="b" />`
```

If we auto-inject the `copy` prop, switching between the two conditional rendering branches would have no effect. Using `copy="!children` has the same problem.

One thing we’re considering is figuring out a way to prevent cross template collisions. Claude is telling me we can’t use `key` for this, and that Lit templates already do something like this. Investigating.

PR #385, opened 2026-08-28
Replace bare Function types with real signatures

brainkim:
Removes every use of the bare `Function` type from src/, per the new house style (`@typescript-eslint/no-unsafe-function-type` in @b9g/eslint-config). `Function` permits calls with any arguments and returns `any`, so none of these call sites were being checked.

- The `schedule`/`cleanup`/`after` callback sets store `(value: any) => unknown` (a `ValueCallback` alias). The parameter is `any` because the module-level maps are keyed by non-generic `ContextState`, erasing each context's `TResult` — but arity and the return type are now checked at the flush sites.
- The pull-controller `onDiff` resolver states its real type, `(diff: Promise<undefined> | undefined) => void`.
- `Retainer.onNextDiff` is `(diff: unknown) => void`: the argument is only used for thenable adoption, and one call site resolves with an array of diffs rather than a promise.
- The enqueue `resolve` is `(value: Promise<undefined> | undefined) => void`.
- `writablePropertyCache` is keyed by `object` — it's keyed by `element.constructor` and never calls the key.
- `ComponentPropsOrProps<T>` tests `T extends (...args: any[]) => unknown` instead of `T extends Function`, which class constructors also matched despite never being components.

Types only; no runtime change. Full suite green on chromium and webkit, `tsc --strict` clean.

### DIFF
--- a/src/crank.ts
+++ b/src/crank.ts
@@ -2304,7 +2304,10 @@ const scheduleMap = new WeakMap<ContextState, Set<ValueCallback>>();
 const cleanupMap = new WeakMap<ContextState, Set<ValueCallback>>();
 
 // keys are roots
-const afterMapByRoot = new WeakMap<object, Map<ContextState, Set<ValueCallback>>>();
+const afterMapByRoot = new WeakMap<
+	object,
+	Map<ContextState, Set<ValueCallback>>
+>();
 
 interface PullController {
 	iterationP: Promise<ChildrenIteratorResult> | undefined;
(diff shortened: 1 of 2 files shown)


src/crank.ts
```diff
@@ -2292,12 +2294,20 @@ function unmountChildren<
 }
 const provisionMaps = new WeakMap<ContextState, Map<unknown, unknown>>();
 
-const scheduleMap = new WeakMap<ContextState, Set<Function>>();
+// The value parameter is any because these sets erase each context's TResult:
+// callbacks are registered as (value: TResult) => unknown but stored keyed by
+// non-generic ContextState.
+type ValueCallback = (value: any) => unknown;
```
brainkim, line 2300: @claude okay so what’s the point of this over Function?

PR #387, opened 2026-09-01
Improve console warning messages with `JSON.stringify`

yyny:
Avoids awkward messages like `but found  while hydrating:` by quoting strings properly.


brainkim:
Shipped in https://github.com/bikeshaving/crank/releases/tag/v0.7.12. Thank you!

